PR #4, opened 2017-09-24
Add Unity3D support.

arodus:


arodus:
#116

arodus:
@matkoch Ready for review, will squash when there are no more issues ;)

PR #13, opened 2018-01-23
Prefer files that have extensions over extensionless files in Windows

GeorgDangl:
When trying to run npm tasks in Windows, I got Win32 Exceptions.
This was because the `ToolPathResolver` found both `npm` and `npm.cmd`. Trying to start a process with the extensionless variant caused the issue.


matkoch:
@GeorgDangl thanks for raising this issue. The fix looks almost okay. The test is probably not. I guess we cannot expect any system to have npm installed. Any suggestion how to test this in a different way? Besides, the test should reside in `Nuke.Common.Tests` :)

GeorgDangl:
Thanks for the Feedback!
I've put the Test in `Nuke.Core.Tests` because I've written it to test the `Nuke.Core.Tooling.ProcessManager`=)

A simple fix could be:

  1. Move the filtering for the executable path to it's own method, e.g.

    public string LocateExecutable(IEnumerable<string> filePaths)
    {
        return filePaths
                    .Where(File.Exists)
                    // In Windows environments, files must have an extension.
                    // This is to prefer, for example, "npm.cmd" over "npm"
                    .Where(x => !EnvironmentInfo.IsWin || Path.HasExtension(x))
                    .FirstOrDefault();
    }
  2. Since there's a hard dependency on `System.IO.File` (which is fine I think), the test could create two files in the `%TEMP%` directory, one with extension and one without.
  3. Move the test to `Nuke.Common.Tests` and test the `ToolPathResolver.LocateExecutable()` instead of the `ProcessManager.StartProcess()`

If that's OK with you, I'll change the code.

matkoch:
I'm not sure how to approach this without letting tests dictate the design of the implementation =/ I'm tempted to just add the fix. This class didn't have a test before. Not sure, need to think =)

PS: greetings from munich to munich! 😁

GeorgDangl:
You could also refactor this into it's own implementation and introduce an `IEnvironmentResolver` or something like that for the `ToolpathResolver`, then it can be mocked for the unit test:

```
var locateExecutable = EnvironmentInfo.IsWin
    ? @"C:\Windows\System32\where.exe"
    : "/usr/bin/which";
var locateProcess = ProcessTasks.StartProcess(
    locateExecutable,
    pathExecutable,
    redirectOutput: true);
locateProcess.AssertWaitForExit();
```

But the naming isn't great yet...

PS: Servus=)

matkoch:

PR #20, opened 2018-02-17
ProcessExtensions are now returning the process.

arodus:
The `ProcessExtensions` are now returning the `IProcess` object to reduce some boilerplate code.


matkoch:
Thanks for your contribution 😄🎉

Issue #30, opened 2018-03-01
ANSI color support


matkoch:
Some console hosts and build servers support ANSI color output. One thing to consider is _when_ to enable them. On Windows systems, ANSI colors usually don't work, whereas UNIX terminals support them.

arodus:
My idea is to check for the `TERM` environment variable (according to https://stackoverflow.com/a/39033815). If the variable equals `xterm-265color` the ansi console sink will be used.

matkoch:
FYI: TeamCity does support ANSI colors, but doesn't set the `TERM` environment variable.

LordMike:
Running stuff in docker could also pose an issue. The container may set `TERM` (or whatever), but the CLI following along (e.g. Powershell, bash, some CI system) might not.

arodus:
You are right, do you have any idea how get the information reliable?

LordMike:
It's probably one long series of guesses that is necessary to get this information.

Nuke already detects certain CI systems / containers - it detected Gitlab for us. This should f.ex. use ANSI colours because Gitlab CI will render it with colours. If the invoking script is powershell, Nuke should use whatever powershell supports.

Running in containers, I don't think it is possible to figure out what the scheme should be - since any number of clients may follow along - but it is at least possible to provide a command switch to override any colour scheme.

E.g. `./build.sh --colors ANSI`

In other news, Gitlab (and others) also support a form of "block" or timing capability for logs. https://gitlab.com/gitlab-org/gitlab-ce/issues/14664#note_56006266

If one writes out the following, and end-marker is made at this unix timestamp: `Console.WriteLine("section_end:1516815288:test\r" + (char)0x1B + "[0K");`

PR #33, opened 2018-03-11
Add ANSI color formating.

arodus:
Still looking for a good condition when to use ANSI formating. #30


arodus:
ANSI coloring is enabled if the `TERM` environment variable starts with `xterm`

PR #47, opened 2018-04-14
Make it possible to use reserved keywords as properties in the nuke metadata definition.

arodus:
- Add escaping of the parameter name in generated extension if the name equals a reserved C# keyword.
- Add appending of an underscore if an enumeration value equals a reserved C# keyword.


matkoch:

PR #49, opened 2018-04-16
Replace NuGet.Client with NuGet.Packaging

kysel:
I've hit dependency hell with NuGet package [NuGet.Client](https://www.nuget.org/packages/NuGet.Client/4.0.0) on which is Nuke.Common dependent and which is not really updated.
I've replaced the NuGet.Client package with the [NuGet.Packaging](https://www.nuget.org/packages/NuGet.Packaging/4.5.0), which is a subset of the NuGet.Client, locally and it seems to work (tests passed, publish nuke to private repo passed and dependency hell resolved).


matkoch:

PR #53, opened 2018-04-19
Move specifications.

matkoch:


matkoch:
There is this line in every generated file:

`// Generated from https://github.com/nuke-build/nuke/blob/specifications/build/specifications/CoverallsNet.json.`

It's using the current branch. Should we make this a fix reference to `master` instead (if current branch is not `develop`)?

Would need to adapt [this](https://github.com/nuke-build/nuke/blob/4f92b7b35a33b237d2a03fb51c6c4153c6d541b1/build/Build.cs#L209) to this:

```
GenerateCode(
    SpecificationsDirectory,
    GenerationDirectory,
    baseNamespace: "Nuke.Common.Tools",
    useNestedNamespaces: true,
    gitRepository: GitRepository.FromLocalDirectory(
        GitRepository.LocalDirectory,
        branch: GitRepository.Branch == "master" ? "master" : "develop"));
```

matkoch:
This would also imply, that we need regenerate whenever publishing from master. Could, however, be combined with the finalization of changelog.

arodus (approved):

PR #63, opened 2018-05-05
Build scripts

### LINKED ISSUE #29: Option in setup script to use local installation of nuget/dotnet
Currently, the bootstrapping scripts always downloads the nuget/dotnet CLT. Although this is more reliable and ensures determinism, it is not always recommendable.

In both setup scripts, there should be added an option to chose between:

1. Always download nuget/dotnet CLT
2. Use local nuget/dotnet installation

This should influence the value of `NuGetFile`/`DotNetFile`. Also the file existence check must be adapted.

matkoch:
Partially fixes #29.

### DIFF
--- a/bootstrapping/build.netcore.ps1
+++ b/bootstrapping/build.netcore.ps1
@@ -1,6 +1,6 @@
 [CmdletBinding()]
 Param(
-    [switch]$NoInit,
+    [switch]$Local,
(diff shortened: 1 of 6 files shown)


matkoch:
Now completely fixes #29.

arodus (approved):

PR #72, opened 2018-06-06
When parsing git repository URLs allow for http URLs

MrWolfZ:
Hi, first of all, thank you for building this great tool.

While looking at the build output on our build server I noticed a warning about a URL that cannot be parsed. Upon inspection of the Nuke code I found that the issue is that we are accessing the git repository via http instead of https. In addition the URL in the warning had a leading whitespace since the whitespace is only trimmed just before applying the regexes.

This PR fixes both issues by making the _s_ in _https_ optional and trimming the whitespace immediately after extracting the URL from the git config file.

Sadly I did not see any way to add tests for this change.


matkoch:
Thanks for your kind words and contribution! 😄🎉

PR #98, opened 2018-07-06
Add support for specify namespaces for XmlTasks

werwolfby:
To be able to do XmlPoke (and other xml operations) we need specify somehow namespaces. It can be done next:

```c#
XmlPoke(xmlPath, "//ns:Element/@Value", "NewValue", ("ns", "http://schemas.xml/2018/ns"));
```

Using power of `ValueTuple` can really simplify syntax from my point of view

### DIFF
--- a/source/Nuke.Common/IO/XmlTasks.cs
+++ b/source/Nuke.Common/IO/XmlTasks.cs
@@ -16,26 +16,26 @@ namespace Nuke.Common.IO
     [PublicAPI]
     public static class XmlTasks
     {
-        public static IEnumerable<string> XmlPeek(string path, string xpath)
+        public static IEnumerable<string> XmlPeek(string path, string xpath, params (string prefix, string uri)[] namespaces)
         {
-            var (elements, attributes) = GetObjects(XDocument.Load(path), xpath);
+            var (elements, attributes) = GetObjects(XDocument.Load(path), xpath, namespaces);
             ControlFlow.Assert(elements.Count == 0 || attributes.Count == 0, "elements.Count == 0 || attributes.Count == 0");
 
             return elements.Count != 0 ? elements.Select(x => x.Value) : attributes.Select(x => x.Value);
         }
 
         [CanBeNull]
-        public static string XmlPeekSingle(string path, string xpath)
+        public static string XmlPeekSingle(string path, string xpath, params (string prefix, string uri)[] namespaces)
         {
-            var values = XmlPeek(path, xpath).ToList();
+            var values = XmlPeek(path, xpath, namespaces).ToList();
             ControlFlow.Assert(values.Count <= 1, "values.Count <= 1");
             return values.SingleOrDefault();
         }
 
-        public static void XmlPoke(string path, string xpath, object value)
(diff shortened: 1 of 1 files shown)


werwolfby:
@matkoch build was failed on `TeamCity`. But I don't think this was because of my changes:
https://teamcity.jetbrains.com/viewLog.html?buildId=1513894&tab=buildResultsDiv&buildTypeId=matkoch_NukeBuild_Nuke#testNameId4602938054387015517

matkoch:
Hey @werwolfby, thanks for the pull-request! No, this issue is not related to your changes. The only thing left for me is to fix the typo from `prifux` to `prefix` :)

@arodus @GeorgDangl how do you like the value tuple syntax?

werwolfby:
@matkoch `prefux` was funny 😄, will fix it soon.

GeorgDangl:
I'm a fan of `System.ValueTuple`, in my opinion it's fine to use it here!

werwolfby:
@GeorgDangl I thought you’ll like it 😏

werwolfby:
@matkoch typo was fixed, branch was rbased to latest dev, and ready to merge.

matkoch:

Unfortunately, next release might take some time due to vacation.

matkoch:
We might change the value tuple syntax back to dictionary. First, this is currently a blocking issue in our documentation #117, and furthermore, I guess it expresses a different thing. Value tuples allow us to specify the same prefix multiple times, which is just wrong in this context. Are you okay with that @werwolfby ?

werwolfby:
@matkoch, I don't like idea with `Dictionary` instead of params of `ValueTuple`s.
The whole power of `XmlPeek` and `XmlPoke` over `XDocument` directly is simplicity of use. And tuples continue keeping this simplicity:
```csharp
XmlPeek(projectFile, "/d:Project/d:ItemGroup/References[@Include = 'Some.Reference']", ("d", "http://www.namespace.com/2013"));
```
It is much cleaner then:
```csharp
XmlPeek(projectFile, "/d:Project/d:ItemGroup/References[@Include = 'Some.Reference']", new Dictionary<string, string>() { {"d", "http://www.namespace.com/2013"} });
```
Yes `Dictionary` has advantage that you can't specify the same `prefix` twice. But specify the same `uri` in most cases will be wrong as well. And compiler anyway do not check this. And we will have run time exception.
I thought that this code should throw exception:
```csharp
                    foreach (var (prefix, uri) in namespaces)
                    {
                        xmlNamespaceManager.AddNamespace(prefix, uri);
                    }
```
But it doesn't 😞 

Anyway we can add this check in start of the method to do the same check as `Dictionary` do and throw the same exception. But I don't think we should sacrifice usability (see example above). 

What do you think? Or you are not agree that value tuple syntax is more readable?

P.S. I'm already use this syntax even if `XDocument` will be faster, but I like `XmlPoke` more because of readability.

P.S.S. And I think I can fix `docfx`. 

P.S.S.S. By the way I'm going to create a new PR with `XmlRemove`

matkoch:
@werwolfby if you fix docfx, I'm good :)

werwolfby:
@matkoch Give me a few days 😏

PR #114, opened 2018-08-10
Print out asserted value on failure

markusamshove:
When creating my first Build I had problems getting paths passed correctly to the tools, resulting in something like this:

before

Since the failure message didn't help me that much (other than that a File didn't exist), I added some output to determine what path I had constructed.

With this PR, tools will generate validation which output the actual value of the property, resulting in this:

after

Which helped me better debugging my issues :-)


arodus:
Thanks for your contribution! :smile::tada:

PR #123, opened 2018-08-15
Add Squirrel as generated tool

markusamshove:


markusamshove:
I've updated the definitions to add all command line options of Squirrel.

I'm not exactly sure on how to design switches that take multiple arguments, like

```
shortcut-locations=
Comma-separated string of shortcut locations, e.g. 'Desktop,StartMenu'
```

Maybe you can help me @matkoch  :-)

arodus:
@MarkusAmshove you can also use `List<?>` as a type. Take a look at [NuGet.json](https://github.com/nuke-build/nuke/blob/develop/build/specifications/NuGet.json#L265-L271) for an example.

markusamshove:
Thanks @arodus 
In my optinion this should be done now.

The `process-start-args` isn't actually a list, they pass it to a `Process` as a single string argument.

build/specifications/Squirrel.json
```diff
@@ -0,0 +1,175 @@
...
+          {
+            "name": "Uninstall",
+            "type": "bool",
+            "format": "--uninstall",
+            "help": "Uninstall the app the same dir as Update.exe."
+          },
+          {
+            "name": "Download",
+            "type": "string",
+            "format": "--download",
```
arodus: Missing `{value}` in format string.

build/specifications/Squirrel.json
```diff
@@ -0,0 +1,175 @@
...
+            "help": "Download the releases specified by the URL and write new results to stdout as JSON."
+          },
+          {
+            "name": "CheckForUpdate",
+            "type": "string",
+            "format": "--checkForUpdate",
```
arodus: Missing `{value}` in format string.

build/Build.GitFlow.cs
```diff
@@ -23,38 +23,34 @@ partial class Build
             }
             else
             {
-                ReleaseFrom($"release/{GitVersion.MajorMinorPatch}");
```
arodus: Rebase onto develop please.

arodus (approved):

arodus (changes requested):
Please rebase onto develop and fix the issues.

build/specifications/Squirrel.json
markusamshove: Those don't take parameters, so I guess "properties" is the wrong category here.
Is there an action in another tool that I can take as example which adds a parameter to the tool, but doesn't take an argument itself?

build/Build.GitFlow.cs
markusamshove: I've resolved the conflicts

arodus (approved):

PR #131, opened 2018-08-25
Add coverlet integration

### LINKED ISSUE #115: Possibility to add coverlet integration
I'm playing with nuke right now and would be awesome to use tools that run on dotnet core ootb.

Coverlet looks promising...
https://github.com/tonerdo/coverlet

Another possible way is to enable custom parameters on DotNetTest wrapper. 
What you guys think about this integration?

ElijahReva:
Resolves: https://github.com/nuke-build/nuke/issues/115


arodus:
Thanks for your contribution,  unfortunately, after discussing this internally, we think we can’t integrate your PR as is for the following reason.

@lucascebertin is already working on an addon for coverlet which uses the coverlet global tool which offers a better integrations.

But the missing possibility to easily set arguments on dotnet test is indeed a problem, 
we could solve this by adding an additional property of the Type `Dictionary<string, object>` to the `DotNetTestSettings`. (#132)

@ElijahReva you're also welcome to join our [Slack Channel](https://slofile.com/slack/nukebuildnet) which reduces the risk of duplicate work.

Issue #148, opened 2018-09-19
Add documentation on how to generate the specs


LordMike:
I wanted to add a spec for Pnpm (a faster version of npm), so I initially just copied Npm.json and set out to find the generator. I did not succeed.

For now, I've copied the generated code and renamed all methods / classes appropriately, and that works - but ideally it should be generated :).

Is pnpm of interest to Nuke in general? (should I do a PR with pnpm.json?)

matkoch:
Given it has 4k stars on GitHub, I think it’s quite reasonable to have it included. A PR would surely be very welcome. However, I suggest to have a quick chat on either Gitter or Slack. We do know that documentation misses a few things, and it’s on our list. But it will take some time to get to that.

matkoch:
Ah sorry, now I understand :)

You just invoke `nuke generate` to execute the [target](https://github.com/nuke-build/nuke/blob/develop/build/Build.CodeGeneration.cs#L31).

Could you please also update your PR for dotnet? (If not I will do it as soon as 0.9.0 is going to ship)

matkoch:
Infrastructure will change fundamentally. So this will be obsolete anyways.

dperez3:
So what is the current way to generate code? The latest version of `CodeGenerator` has an API that's different from all the documentation I can find.

Where do I get the json schema for json spec files?

matkoch:
Schema is in the examples, and code gets generated on compile.

PR #151, opened 2018-09-24
WIP: Copied npm.json to pnpm.json

LordMike:
See #148

I've copied the npm.json and changed the references to pnpm to make it work. This PR still needs a number of changes, f.ex. verify that all commands are in place and the descriptions are correct. The references back to some form of source documentation needs to be fixed.

### DIFF
--- a/build/specifications/Pnpm.json
+++ b/build/specifications/Pnpm.json
@@ -1,0 +1,131 @@
+{
+  "$schema": "https://raw.githubusercontent.com/nuke-build/nuke/master/source/Nuke.CodeGeneration/schema.json",
+  "references": [
+    "https://docs.npmjs.com/cli/install#//div[@id='page']",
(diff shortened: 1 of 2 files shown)


matkoch:
I'm against blindly copying the specification from npm. If the set of commands is exactly as npm, then I suggest to introduce a flag `NpmTasks.UsePnpm` or something. If it's not the same set, then at least the help text of the task should explain a little more, what pnpm does in comparison to npm. Usually we try to find official texts for that, which are just copied.

matkoch:
@LordMike current version allows call `SetVariable("NPM_EXE", toolpath)` and it will be picked up automatically for `NpmTasks`. This basically works for all tool tasks (https://github.com/nuke-build/nuke/commit/d42da72c28f79abe117659248d28e0c2a54b6fbc).

I will close the PR. If you have any further feedback, feel free to reopen.

PR #157, opened 2018-10-20
Add tool delegate

arodus:


source/Nuke.Common/Tooling/PackageExecutableAttribute.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            _packageId = packageId;
+            _packageExecutable = packageExecutable;
+            _framework = framework;
+        }
+
+        public override object GetValue(string memberName, Type memberType)
+        {
+            var toolPath =
+                ToolPathResolver.TryGetEnvironmentExecutable($"{memberName.ToUpperInvariant()}_EXE") ??
```
matkoch: I think this is worth to discuss. For `PathExecutable` it kind of makes sense, but what about executables from NuGet packages? For instance, for `Xunit` the executable is called `xunit.console.exe`. Should we still allow to override via `XUNIT_EXE` ?

source/Nuke.Common/Tooling/PackageExecutableAttribute.cs
```diff
@@ -0,0 +1,55 @@
...
+    ///     });
+    ///     </code>
+    /// </example>
+    public class PackageExecutableAttribute : InjectionAttributeBase
+    {
+        private readonly string _packageId;
+        private readonly string _packageExecutable;
+        private readonly string _framework;
+
+        public PackageExecutableAttribute(string packageId, string packageExecutable, string framework = null)
```
matkoch: Make `framework` a property and instead have two overloads `PEA(packageId, packageExecutable)` and `PEA(packageId, packageExecutable32, packageExecutable64)` which allows better support for 32/64 bit executable distinction by checking `EnvironmentInfo.Is32Bit`. Same for `PathExecutableAttribute` of course.

PR #159, opened 2018-10-23
Pass build type to injection attributes

arodus:


source/Nuke.Common/Execution/InjectionAttributeBase.cs
```diff
@@ -22,6 +23,13 @@ namespace Nuke.Common.Execution
     public abstract class InjectionAttributeBase : Attribute
     {
         [CanBeNull]
-        public abstract object GetValue(string memberName, Type memberType);
+        [Obsolete("Use GetValue(string memberName, Type memberType, MemberInfo memberInfo) instead.", error: true)]
```
matkoch: It's probably not called from user code, so won't have any effect. Imo, just remove.

source/Nuke.Common/Execution/InjectionService.cs
```diff
@@ -17,6 +17,7 @@ public static void InjectValues(NukeBuild build)
             var injectionMembers = build.GetInjectionMembers()
                 .OrderByDescending(x => x.GetCustomAttribute<ParameterAttribute>() != null);
 
+            var buildType = build.GetType();
```
matkoch: Would prefer inlining.

PR #160, opened 2018-10-24
Allow build project to be packed as global tool

robsonj:


robsonj:
Changes to allow a Nuke based build to be installed and run as a global tool.
Changes are work from matkoch and tweaks from myself to get an initial implementation working

PR #164, opened 2018-11-04
Add executable search when using Paket

Liminiens:
Hi, I recently tried nuke with [paket](https://fsprojects.github.io/Paket/) manager and it seems like there is an issue with executable resoultion when using paket (`GitVersion.exe` from nuke install wizard in my case) . This adds search in paket packages folder if paket is installed. Paket stores packages like `packages/<packageId>/..`. I tried to test the code myself on test solution, but I couldn't find the correct way to test the local nuke build without errors :) Though, during debug exe path lookup worked correctly.

image

### DIFF
--- a/source/Nuke.Common.Tests/PathConstructionTest.cs
+++ b/source/Nuke.Common.Tests/PathConstructionTest.cs
@@ -22,6 +22,7 @@ namespace Nuke.Common.Tests
         public void TestParent(string path, string expected)
         {
             ((AbsolutePath) path).Parent.Should().Be((AbsolutePath) expected);
+            ((string) ((AbsolutePath) path).Parent).Should().Be(expected);
         }
         
         [Theory]
--- a/source/Nuke.Common/IO/PathConstruction.cs
+++ b/source/Nuke.Common/IO/PathConstruction.cs
@@ -266,17 +266,20 @@ namespace Nuke.Common.IO
 
             public static explicit operator RelativePath([CanBeNull] string path)
             {
+                if (path is null)
+                    return null;
+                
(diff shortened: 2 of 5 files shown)


matkoch:
Hi Slava,

thanks for your contribution! It's looking good. I still would like to understand, in which case Paket would be preferred. It is my understanding, that with Paket the packages would be located within the repository, whereas with NuGet (package references), they are located in a machine-wide folder, making reuse much easier and storage footprint most efficient.

Liminiens:
Yes, paket stores them locally and resolves them from that folder. However, as I have found out now, there is also [storage: none](https://fsprojects.github.io/Paket//dependencies-file.html#Disable-packages-folder) option which removes `packages` usage and falls back to nuget cache. Also there are [package groups](https://fsprojects.github.io/Paket/groups.html) and they exist in form of subfolders in `packages` like `packages/<groupName>/<packageId>`.

Well, I tried `storage: none` option and seems like there are still some quirks how paket works with nuget as it doesnt solve the problem above :/

Also after that new problem emerged with `Glob.Extensions`

image

where the clean task defined as:

```
DeleteDirectories(GlobDirectories(SourceDirectory, "**/bin", "**/obj"));
DeleteDirectories(GlobDirectories(TestsDirectory, "**/bin", "**/obj"));
EnsureCleanDirectory(OutputDirectory);
```

So, I think that I will work on some correct repro of the problem and look into [Paket.Core](https://www.nuget.org/packages/Paket.Core/) to find some public api which maybe will help with the problem above.

Liminiens:
I updated logic to also search in all subdirectories if paket groups are used

matkoch:

PR #175, opened 2018-11-18
Change build.ps1 behaviour to use exit rather than SetShouldExit

### LINKED ISSUE #173: build.ps1, Ctrl+C closes window
It seems the Nuke build.ps1 closes the Powershell window when canceled (Ctrl+C).

For example:

* In a new PS window, running a command and cancelling **does not** close the window
* In a new PS window, running `.\build.ps1` and cancelling **does** close the window

This is minorly inconvenient, as I will iterate on builds quickly a number of times, meaning I'd run the build script, do a change, cancel the build, re-do the build etc.. But the cancelling step means I have to reopen my PS window :P

LordMike:
Fixes #173 

Uses `exit` rather than `$host.SetShouldExit(1)` to set the exit code. Given [this SO answer](https://stackoverflow.com/questions/4391553/how-to-return-an-exit-code-from-a-powershell-script-only-when-run-non-interactiv), I gather one should use `exit 1` rather than `$host.SetShouldExit(1)`. 

I've run a modified build.ps1 on a project of mine, and found that if I use `exit`, the `$LASTEXITCODE` is `0` when the build completes, and `1` when I cancel it with Ctrl+C. The shell does _not_ close when using `exit`.

### DIFF
--- a/source/Nuke.GlobalTool/templates/build.netcore.ps1
+++ b/source/Nuke.GlobalTool/templates/build.netcore.ps1
@@ -7,7 +7,7 @@ Param(
 
 Write-Output "Windows PowerShell $($Host.Version)"
 
-Set-StrictMode -Version 2.0; $ErrorActionPreference = "Stop"; $ConfirmPreference = "None"; trap { $host.SetShouldExit(1) }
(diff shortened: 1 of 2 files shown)


matkoch:
Looks good. Thanks for your contribution!

PR #176, opened 2018-11-18
Add linux replacement of $-variables

### LINKED ISSUE #165: Expand variables on Unix
https://github.com/nuke-build/nuke/blob/c209e6ad6057f70d735f0427fd5eccd217f96426/source/Nuke.Common/EnvironmentInfo.Others.cs#L28-L32

LordMike:
Fixes #165 

Uses a regex of `\$([a-z0-9_\-]+)` to locate `$VARIABLES` and replace them using regex's replacer.

### DIFF
--- a/source/Nuke.Common/EnvironmentInfo.Others.cs
+++ b/source/Nuke.Common/EnvironmentInfo.Others.cs
@@ -31,7 +31,9 @@ namespace Nuke.Common
             string ExpandUnixEnvironmentVariables()
             {
                 var regex = new Regex(@"\$([a-z_][a-z0-9_]*)", RegexOptions.IgnoreCase);
-                return regex.Replace(value, x => Environment.GetEnvironmentVariable(x.Groups[1].Value));
+                return regex.Replace(
+                    value.ReplaceRegex("^~", x => Environment.GetEnvironmentVariable("HOME")),
+                    x => Environment.GetEnvironmentVariable(x.Groups[1].Value));
             }
 
             return IsWin
--- a/source/Nuke.Common/EnvironmentInfo.Others.cs
+++ b/source/Nuke.Common/EnvironmentInfo.Others.cs
@@ -30,10 +30,9 @@ namespace Nuke.Common
         {
             string ExpandUnixEnvironmentVariables()
             {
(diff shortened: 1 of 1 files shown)


LordMike:
Is it possible to write unit tests for the different platforms?

LordMike:
Tested. Run by hacking the Environment to simulate Linux/Windows.

```csharp
            Console.WriteLine("Win: " + EnvironmentInfo.ExpandVariables("%SystemDrive%"));
            Console.WriteLine("Únx: " + EnvironmentInfo.ExpandVariables("$SystemDrive"));
```

```
As Windows:
Win: C:
Unx: $SystemDrive

As Linux:
Win: C:
Unx: C:
```

LordMike:
So. I ran it on linux in WSL, and found that .NET's `Environment.GetEnvironmentVariable()`  is case-sensitive.  If this is the intended behaviour, then all is ok. Else the code needs to have its Environment lookup thingy tweaked.

LordMike:
Fixed regex to `\$([a-z_][a-z0-9_]*)`, given this [SO answer](https://stackoverflow.com/a/2821183).

matkoch:
I force-pushed a draft how I would like the method to look like. One thing that's missing, is expansion for `~` to the home directory. Only at the beginning of the string.

LordMike:
I've added the expansion

LordMike:
Why not use `ReplaceRegex` for both replacements?

matkoch:
How exactly? Feel free to commit.

LordMike:
One could even interleave the two replacements (doing a regex like `(^~| ... )` and then doing an if in the evaluator) :P

matkoch:
Looks better, thanks! But why ignore case? I think that should be removed. I will do it myself when merging, if you agree.

LordMike:
Well. That, or change the regex to `\$([a-zA-Z_][a-zA-Z0-9_]*)`.

matkoch:
Ah sorry, I confused it with replacing the variable matching case-insensitive. I guess it's correct like it is.

matkoch:
Thanks for your contribution :)

Issue #178, opened 2018-11-20
build.ps1 failure if sdk is not present in global.json


robsonj:
We have a situation where we're using the global.json file for other SDK versions, but not the .Net Core SDK version.

Our global.json file looks like...
```
{
  "msbuild-sdks": {
    "MSBuild.Sdk.Extras": "1.6.61",
    "Microsoft.Build.CentralPackageVersions": "2.0.1"
  }
}
```

This causes the build.ps1 script to fail as it is expecting to find the SDK JSON entry in the global.json file when it finds a global.json file.

PR #179, opened 2018-11-20
Fix for Issue #178

robsonj:


matkoch:
I think there is a better way of checking if a property exists instead of deliberately working with exceptions.

matkoch:
Also the bash script must be adapted too.

robsonj:
Fine to improve it further. My immediate issue was that all of our builds began failing as soon as the global.json was added. This at least resolved that issue.

PR #269, opened 2019-04-23
Fix creation of solution file as slashes are not allowed in folder names

kthompson:


kthompson:
I am not sure if this is just a Visual Studio restriction but when I tried to load the solution I was met with these errors:

```
nuke-build/resharper : error  : Solution Folder names cannot:
- contain any of the following characters: / ? : & \ * " < > | # %
- contain Unicode control characters
- contain surrogate characters
- be system reserved names, including 'CON', 'AUX', 'PRN', 'COM1' or 'LPT2'
- be '.' or '..'```

PR #340, opened 2019-11-16
Added code cleanup from `JetBrains.ReSharper.CommandLineTools`

david-driscoll:
https://www.jetbrains.com/help/resharper/CleanupCode.html


david-driscoll:
@matkoch based on your question in slack.

1) Yes this was copied from `InspectCode`.
2) The only change between `InspectCode` and this is the removal of the `PostProcess` step since there is no data that needs to be reported to TeamCity (that I'm aware of).

PR #369, opened 2020-01-03
Treat an empty CI-specific environment variable as non-existing

dennisdoomen:
When running Nuke from inside a LINUX docker container, we tend to pass the TEAMCITY_VERSION from the host to the container. However, that will set the value to an empty string, even though the variable did not exist on the host. This is fixed for Team City and all the other CI providers.

There were no tests that I could update/amend

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common/CI/AppVeyor/AppVeyor.cs
+++ b/source/Nuke.Common/CI/AppVeyor/AppVeyor.cs
@@ -40,7 +40,7 @@ namespace Nuke.Common.CI.AppVeyor
 
         public static AppVeyor Instance => NukeBuild.Host == HostType.AppVeyor ? s_instance.Value : null;
 
-        internal static bool IsRunningAppVeyor => Environment.GetEnvironmentVariable("APPVEYOR") != null;
+        internal static bool IsRunningAppVeyor => !Environment.GetEnvironmentVariable("APPVEYOR").IsNullOrEmpty();
 
         internal AppVeyor()
         {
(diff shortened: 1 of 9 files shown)


dennisdoomen:
@matkoch any idea what's going on with the builds?

matkoch:
Nothing related to your changes

dennisdoomen:
In that case it would be nice to get a pre-release with these changes.

matkoch:
@dennisdoomen are the environment variables propagated by default, or are you doing this for some other reason? I'm asking because in case of the first, we should make the same changes to all CI server interfaces. In the latter case, the better option would be to pass `--host console`.

dennisdoomen:
We do something like this:

```
docker run --rm --name $env:build_container_name `
 -v /var/run/docker.sock:/var/run/docker.sock `
 -v $PWD/$env:artifacts_dir`:/repo/$env:artifacts_dir `
 -v $PWD/.git:/.git `
 --network host `
 -e TEAMCITY_VERSION=$env:TEAMCITY_VERSION  `
```

So even if the `TEAMCITY_VERSION` env variable wasn't set, the docker container will see it as `""`. 

So yes, most likely it will also affect the other build systems in a similar way.

matkoch:
But why are you passing the TeamCity version like this?

dennisdoomen:
How else does Nuke know that the build is running on TeamCity

dennisdoomen:
So? Are you accepting my PR?

matkoch:
We can do it like that. But could you apply the same logic to all other CI
interfaces?

dennisdoomen:
Sure, will do.

cmenzi:
When an empty envvar is treated as not existent, how to deal with switch only parameters?

For example if you have a - - force parameter and a env var FORCE should do the same?

Or do this only apply on CI related env vars?

matkoch:
`envvar=1` … besides, we’re only talking about CI system detection here.
Not general parameter resolution.

PR #389, opened 2020-02-02
UnityTasks improvements

razenpok:
This PR:

- Synchronizes the path resolving code making it more consistent. This includes throwing an exception in `GetToolPathViaManualInstallation` on unsupported platforms like `GetToolPathViaHubVersion` does
- Improves assertion warning message when `ProjectPath` is not set, giving the user a hint on a possible fix

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common/Tools/Unity/UnityTasks.cs
+++ b/source/Nuke.Common/Tools/Unity/UnityTasks.cs
@@ -35,31 +35,28 @@ namespace Nuke.Common.Tools.Unity
         {
             return EnvironmentInfo.Platform switch
             {
-                PlatformFamily.Windows => $@"{EnvironmentInfo.SpecialFolder(
-                    EnvironmentInfo.Is32Bit
-                        ? SpecialFolders.ProgramFilesX86
-                        : SpecialFolders.ProgramFiles)}\Unity\Editor\Unity.exe",
+                PlatformFamily.Windows => $@"{GetProgramFiles()}\Unity\Editor\Unity.exe",
                 PlatformFamily.OSX => "/Applications/Unity/Unity.app/Contents/MacOS/Unity",
-                PlatformFamily.Linux => null,
-                PlatformFamily.Unknown => null,
                 _ => null
             };
         }
 
         private static string GetToolPathViaHubVersion(string version)
         {
-            static string GetProgramFiles()
-                => EnvironmentInfo.SpecialFolder(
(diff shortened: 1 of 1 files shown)


source/Nuke.Common/Tools/Unity/UnityTasks.cs
```diff
@@ -35,36 +35,36 @@ private static string GetToolPathViaManualInstallation()
...
-                PlatformFamily.Windows => $@"{EnvironmentInfo.SpecialFolder(
-                    EnvironmentInfo.Is32Bit
-                        ? SpecialFolders.ProgramFilesX86
-                        : SpecialFolders.ProgramFiles)}\Unity\Editor\Unity.exe",
+                PlatformFamily.Windows => $@"{GetProgramFiles()}\Unity\Editor\Unity.exe",
                 PlatformFamily.OSX => "/Applications/Unity/Unity.app/Contents/MacOS/Unity",
-                PlatformFamily.Linux => null,
-                PlatformFamily.Unknown => null,
-                _ => null
+                _ => throw new Exception($"Cannot determine Unity installation path for current platform.")
```
matkoch: This method should really return `null`. The generic exception will be thrown [here](https://github.com/matkoch/common/blob/develop/source/Nuke.Common/Tooling/ProcessTasks.cs#L52).

razenpok: Is there a way to throw an exception with a more meaningful message? "ToolPath was not set." doesn't seem too helpful, since it doesn't provide any details (which is the platform being unsupported in this particular case).

https://blog.ploeh.dk/2014/12/23/exception-messages-are-for-programmers/

matkoch: I expect developers being able to work with a stacktrace.

razenpok: Ok, you've got a point. "ToolPath was not set." will cause them to set it manually anyway, so it's not that big of a deal.

Should I change `GetToolPathViaHubVersion` to return `null` as well?

matkoch: I think no. In this case the user was using `SetHubVersion`  .. and if that doesn't resolve, it's clearly worth a custom exception.

razenpok: Good point. Pushed a fix.

Issue #346, opened 2019-11-20
TeamCity env variables parser error


MichaelSL:
### Description
Build is failing on Team City when trying to parse env variables. I believe this is because I have several `=` in the variable

### Steps to reproduce
I have several variables with `=` in the name in TC:

```

dep.BuildBootstrapper.env.=:: = ::\
dep.BuildBootstrapper.env.=C: = C:\BuildAgent\bin
dep.BuildBootstrapper.env.=ExitCode = 00000000
```

### Relevant log output

```
[15:11:10]
[Step 1/2] An item with the same key has already been added. Key: dep.BuildBootstrapper.env.
[15:11:10]
[Step 1/2]    at System.Collections.Generic.Dictionary`2.TryInsert(TKey key, TValue value, InsertionBehavior behavior)
[15:11:10]
[Step 1/2]    at Nuke.Common.CI.TeamCity.TeamCity.ParseDictionary(String file)
[15:11:10]
[Step 1/2]    at Nuke.Common.CI.TeamCity.TeamCity.<.ctor>b__14_1()
[15:11:10]
[Step 1/2]    at System.Lazy`1.ViaFactory(LazyThreadSafetyMode mode)
[15:11:10]
[Step 1/2]    at System.Lazy`1.ExecutionAndPublication(LazyHelper executionAndPublication, Boolean useDefaultConstructor)
[15:11:10]
[Step 1/2]    at System.Lazy`1.CreateValue()
[15:11:10]
[Step 1/2]    at System.Lazy`1.get_Value()
[15:11:10]
[Step 1/2]    at Nuke.Common.CI.TeamCity.TeamCity.get_ConfigurationProperties()
[15:11:10]
[Step 1/2]    at Nuke.Common.CI.TeamCity.TeamCity.get_BranchName()
[15:11:10]
[Step 1/2]    at Nuke.Common.Git.GitRepositoryAttribute.GetBranch()
[15:11:10]
[Step 1/2]    at Nuke.Common.Git.GitRepositoryAttribute.GetValue(MemberInfo member, Object instance)
[15:11:10]
[Step 1/2]    at Nuke.Common.Execution.InjectionUtility.InjectValuesInternal[T](T instance, IEnumerable`1 tuples)
[15:11:10]
[Step 1/2]    at Nuke.Common.Execution.InjectionUtility.InjectValues[T](T instance, Func`2 filter)
[15:11:10]
[...]

PR #392, opened 2020-02-03
Fix TeamCity env variables parser error #346

ghost:

- [X] Is based on my own work

### DIFF
--- a/source/Nuke.Common/Utilities/String.IndexOf.cs
+++ b/source/Nuke.Common/Utilities/String.IndexOf.cs
@@ -13,7 +13,7 @@ namespace Nuke.Common.Utilities
         public static int IndexOfRegex(this string text, string expression)
         {
             var regex = new Regex(expression, RegexOptions.Compiled);
-            return regex.Match(text).Index + 1;
+            return regex.Match(text).Index;
         }
     }
 }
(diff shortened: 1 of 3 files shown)


source/Nuke.Common/Utilities/String.IndexOf.cs
```diff
@@ -0,0 +1,19 @@
...
+
+namespace Nuke.Common.Utilities
+{
+    public static partial class StringExtensions
+    {
+        [Pure]
+        public static int IndexOfRegex(this string text, string expression)
+        {
+            var regex = new Regex(expression, RegexOptions.Compiled);
+            return regex.Match(text).Index + 1;
```
matkoch: I'm curious why `index + 1` ?

ghost: So it really depends on the regular expression, however for the one that teamcity uses it actually matches the character before the last = sign, so the +1 moves the index to the actual position of the equal sign. I think it's probably better if this method doesn't do the +1, and instead have the caller worry about such things. I will make a small update to the PR.

matkoch: That's what I thought. Thanks!

matkoch:

matkoch:
I will release a 0.24.1 asap.

PR #406, opened 2020-02-15
Add support for Nerdbank GitVersioning

macsux:

- [ X] Follows the contribution guidelines
- [ X] Is based on my own work
- [ X] Is in compliance with my employer

### DIFF
--- a/build/specifications/NerdbankGitVersioning.json
+++ b/build/specifications/NerdbankGitVersioning.json
@@ -1,0 +1,237 @@
+{
+  "$schema": "https://raw.githubusercontent.com/nuke-build/nuke/master/source/Nuke.CodeGeneration/schema.json",
+  "references": [
+    "https://github.com/AArnott/Nerdbank.GitVersioning/blob/master/doc/nbgv-cli.md"
+  ],
+  "name": "NerdbankGitVersioning",
+  "officialUrl": "https://github.com/AArnott/Nerdbank.GitVersioning",
+  "packageId": "nbgv",
+  "packageExecutable": "nbgv.dll",
+  "tasks": [
+    {
+      "help": "Prepares a project to have version stamps applied using Nerdbank.GitVersioning",
+      "postfix": "Install",
+      "definiteArgument": "install",
+      "settingsClass": {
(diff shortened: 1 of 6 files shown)


tunger:
@matkoch is there something left to do? I'd be interested as well in this feature.

matkoch:
@tunger yes I was also looking at it again, because I wanna switch from GitVersion.

One thing we need to figure out is how the task name should be. `NerdbankGitVersioning` looks a bit long, especially when you consider the attribute:

```
[NerdbankGitVersioning] NerdbankGitVersioning GitVersion;
```

I'm actually leaning more towards

```
[Nbgv] Nbgv Nbgv;
```

What do you think?

GeorgDangl:
Actually, I'd just prefer the long name - then at least I don't have to figure out what `Nbgv` means😀

Also unrelated to the PR, @matkoch, why do you want to switch?

macsux:
This PR has been hanging for a long time.. Any reason it's not getting
merged?

matkoch:
@macsux it's being discussed currently, right? As always the reason is "limited time"...

tunger:
@matkoch I also like verbose names better. It's usually only the declaration that gets this long. `nbgv` is fine for the dotnet global tool as you potentially write that more often, but I don't see a reason to shorten it here.

matkoch:
@tunger, maybe `NerdbankGitVersion` at least? dropping the `ing` for both attribute and tasks?

tunger:
@matkoch well, I'd still go for the same name as the repo, but `NerdbankGitVersion` seems like a good compromise ;) 
I looked at what the maintainer called his packages, mostly he uses the full name ([npm](https://www.npmjs.com/package/nerdbank-gitversioning), [github action](https://github.com/marketplace/actions/nerdbank-gitversioning), [nuget](https://nuget.org/packages/Nerdbank.GitVersioning)), as for Cake he uses [GitVersioning](https://github.com/dotnet/Nerdbank.GitVersioning/blob/master/doc/cake.md). But that's ambigious regarding GitVersion.
It's your call :)

PR #408, opened 2020-02-18
Cache candidate packages

mortengjesing:
After [Fix NuGetPackageResolver to resolve dependencies by default](https://github.com/nuke-build/nuke/commit/039405d5bd34b7ba569c5d32f81091b7b81325d1#diff-ff1356e2ed38bc7bc09f93075cb7c46d) commit, looking up packages has been really slow (minutes in my case)

So I have introduces caching to speed things up.

Test performance.
Before 19.3892537 sec
After 1.4985964 sec
After, 2. lookup 0.0635239 sec

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common.Tests/Nuke.Common.Tests.csproj
+++ b/source/Nuke.Common.Tests/Nuke.Common.Tests.csproj
@@ -5,6 +5,12 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <None Include="Tooling\NuGetPackageResolverCacheTest.csproj">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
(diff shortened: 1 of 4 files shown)


matkoch:
Can you provide a repro?

mortengjesing:
Added a test with cache enabled/disabled for repro

GuillaumeLebeau:
Wouldn't it be interesting to have also a cache at the level of the ToolPathResolver class?
If we have a tool which is call multiple times, each time the `ToolPathResolver.GetPackageExecutable` method is called but the same tools is needed. Overall performance would be better if the first time the path is resolved and cached, then the next time the path is directly returned from cache.

matkoch:
Should be fixed with https://github.com/nuke-build/nuke/milestone/44

Issue #412, opened 2020-02-24
Logging in Jenkins


magicmonty:
**Update: Removed the End of task requests, as this is a different feature, which does not belong here ;)**

### Description
On BuildServers (in my case Jenkins) the output of the Nuke build log is quite ugly and hard to parse:

![grafik](https://user-images.githubusercontent.com/315639/75135565-36310000-56e2-11ea-8d91-3f78c3c6cf6c.png)

and

![grafik](https://user-images.githubusercontent.com/315639/75135673-9a53c400-56e2-11ea-95d8-228a9914348e.png)

It would be nice, if there would be a command line option where one can enable a simpler logging for the Tasks.

Proposal for the Task Start: 

```
---- Starting Task Clean ---- 
```

### Alternatives
Start:
```
+----------
| Clean
+--
```

or 
```
+==========
| Clean
+==
```

Another alternative could be to just implement this as a special output sink for Jenkins

magicmonty:
I will provide a PR for this ;)

magicmonty:
I changed this issue to Logging in Jenkins, because this was in fact two feature requests.

matkoch:
can you try this please? https://medium.com/pacroy/how-to-fix-jenkins-console-log-encoding-issue-on-windows-a1f4b26e0db4

<blockquote><img src="https://miro.medium.com/max/1200/1*unw_OAKk9mV1EEhiNovhGg.jpeg" width="48" align="right"><div><img src="https://cdn-static-1.medium.com/_/fp/icons/favicon-rebrand-medium.3Y6xpZ-0FSdWDnPM3hSBIA.ico" height="14"> Medium</div><div><strong><a href="https://medium.com/pacroy/how-to-fix-jenkins-console-log-encoding-issue-on-windows-a1f4b26e0db4">How to Fix Jenkins Console Log Encoding Issue on Windows</a></strong></div><div>If you get alien characters in console log of your job like this:</div></blockquote>

magicmonty:
Unfortunately this doesn't work for me. I tried both `-Dfile.encoding=UTF8` and `-Dfile.encoding=UTF-8` variants.

magicmonty:
Now it shows 
![grafik](https://user-images.githubusercontent.com/315639/75158026-9b521900-5715-11ea-933c-a3d4cad6bce7.png)

matkoch:
How are you invoking the build?

magicmonty:
[...]

PR #413, opened 2020-02-24
Add a new OutputSink for Jenkins

magicmonty:

- [x] Is based on my own work

This is the PR for #412


matkoch:
I'm sorry for having to say this, but I'm not going to bother with a PR that includes nearly 50% copy/paste without even disclosing it.

magicmonty:
Ok, my first thought was to introduce protected properties for Corners and Edges and overwrite them in the Jenkins OutputSink but I thought this could be a bit too intrusive. Would this be a better option then? 

Also this could be made into a more Generic OutputSink (i.E. AnsiOnlyOutputSink) which could then set in an overridden OutputSink-Property in the Build.cs if this would be better?

magicmonty:
See last commit ;)

PR #427, opened 2020-03-06
Support for default implementations of interfaces

tunger:
This might be an exotic proposal, but let me explain my line of thinking. I am curious about your opinion.

I want to create a library in my company providing targets for different use cases, like uniform versioning or pushing to Octopus. Right now I created a new abstract class combining all these targets, but I feel like this is overblown for some projects where they want to use only parts of the targets I provide.

So I thought, what about the new C#8 default implementations of interfaces? Utilizing them like "traits" where you can add the functionality you want. It turned out there were not many changes necessary in NUKE. Even parameter injection works (however, you can't set them interactively because you can only provide get-properties in interfaces). I needed to add a helper method to be able to reference targets in interfaces (because they are hidden in the implementing type).

You can now do something like this:

```csharp
interface IUseInterface
{
    [Parameter] string InterfaceParameter => EnvironmentInfo.GetParameter<string>(nameof(InterfaceParameter));

    public Target InterfaceTarget => _ => _
           .Requires(() => InterfaceParameter)
           .Executes(() =>
           {
               Logger.Info($"I am using parameter value {InterfaceParameter} ...");
           });
}

partial class Build : NukeBuild, IUseInterface
{
    Target DependOnInterfaceTarget => _ => _
        .DependsOn(FromInterface<IUseInterface>(_ => _.InterfaceTarget))
        .Executes(() => { Logger.Info("I run after the interface target."); });
}
```

What do you think?

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common.Tests/Execution/DefaultInterfaceExecutionTest.cs
+++ b/source/Nuke.Common.Tests/Execution/DefaultInterfaceExecutionTest.cs
@@ -1,0 +1,143 @@
+﻿using System;
+using System.Linq;
+using System.Linq.Expressions;
+using FluentAssertions;
+using Nuke.Common.Execution;
+using Xunit;
+
+namespace Nuke.Common.Tests.Execution
+{
+    public class DefaultInterfaceExecutionTest
+    {
+        public static string Description = "description";
+        public static Action Action = () => { };
+        public static Expression<Func<bool>> Requirement = () => true;
+        public static Expression<Func<bool>> StaticCondition = () => true;
+        public static Expression<Func<bool>> DynamicCondition = () => false;
+
+        [Fact]
+        public void Test()
+        {
+            var build = new TestBuild();
+            var targets = ExecutableTargetFactory.CreateAll(build, x => x.E);
+
+            var a = targets.Single(x => x.Name == nameof(ITestBuild.A));
+            var b = targets.Single(x => x.Name == nameof(ITestBuild.B));
+            var c = targets.Single(x => x.Name == nameof(ITestBuild.C));
+            var d = targets.Single(x => x.Name == nameof(ITestBuild.D));
+            var e = targets.Single(x => x.Name == nameof(TestBuild.E));
+
+            targets.Single(x => x.IsDefault).Should().Be(e);
+
+            a.Description.Should().Be(Description);
+            a.Requirements.Should().Equal(Requirement);
+            a.Actions.Should().Equal(Action);
+            a.AllDependencies.Should().BeEmpty();
+
+            b.DependencyBehavior.Should().Be(DependencyBehavior.Execute);
+            b.StaticConditions.Should().Equal(StaticCondition);
+            b.ExecutionDependencies.Should().Equal(d);
+            b.TriggerDependencies.Should().Equal(c);
+            b.AllDependencies.Should().NotBeEmpty();
+
+            c.Triggers.Should().Equal(b);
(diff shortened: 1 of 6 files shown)


tunger:
Hm I don't know why appveyor ubuntu build fails. I wait for your opinion on this PR before further investigating. You may ignore the netcoreapp3.1 change, thought this might be it.

matkoch:
@tunger great that you took the time to investigate this. Actually, default interface implementations is something I've been looking at already a while ago. It was just a time issue. Chances are high we can merge this pretty soon, but one thing that instantly catches my eye is the additional test project. I don't want another test project. You can raise the target framework for the existing project.

tunger:
Great! Removed the test project, that feels better now :)
AppVeyor still fails. I don't have much experience with that one.
Maybe there should be another test for parameters in default interfaces, but I'm not adept yet with the NUKE internals. Could be that there is an even better solution for parameters than proposed by me (via EnvironmentInfo), but it is usable that way at least.

matkoch:
@tunger if you're interested, I will be discussing this PR next time I'm doing a live stream.

tunger:
@matkoch sorry, was a bit busy lately. On what platform do you stream?

I did some further tests, seems like multiple interface inheritance is no problem as reflection lists all implemented interfaces.
I added overloads for Requirements. It seems it's only possible to make it working properly via readonly properties in interfaces which utilize EnvironmentInfo to get the parameter value.

Any ideas what we further should check/add?

matkoch:
@tunger  I just finished the stream :) Can you contact me on [Slack](https://slofile.com/slack/nukebuildnet) ?

matkoch:
@tunger  I don't think the `Requires` overload makes much sense. If requirements are set, than rather in the interface itself. I'd vote to wait for a proper use case.

matkoch:
And parameters should resolve like this:

```
[Parameter] string StringArgument => InjectionUtility.GetInjectionValue(() => StringArgument);
[Solution] Solution Solution => InjectionUtility.GetInjectionValue(() => Solution);
```

tunger:
> 
> 
> @tunger I don't think the `Requires` overload makes much sense. If requirements are set, than rather in the interface itself. I'd vote to wait for a proper use case.

I was thinking of the following use case. In my company the build server provides server urls, api keys, ... as global variables. An interface could implement such well known parameters.

```
[Parameter] string InterfaceParameter => InjectionUtility.GetInjectionValue(() => InterfaceParameter)
    ?? EnvironmentInfo.GetParameter<string>("buildserver.interfaceParameter");
```

For example, nuget api key, used in publish targets; or version number when using custom versioning.
But I now realised that the overload does not improve things much, because in order to use the value, one needs to cast to the interface anyway. See line 2, it works that way as well.
```
.Requires((ITestBuild x) => x.InterfaceParameter)
.Requires(() => ((ITestBuild)this).InterfaceParameter)
```
You are probably right to drop it.

Thanks for the hint with InjectionUtility.

tunger:
Oh, whoops, thought this was already merged and wanted to use it ;)
Anything still to figure out?

matkoch:
Hi @tunger, afair it's almost ready. There were just too many bug reports recently, that I didn't have any time to invest in a new major release. I will try to check and merge this until next week.

tunger:
@matkoch great, thanks, no hurry! If I can help with anything related to this, please let me know. Couldn't remember if there were any pending issues.

Issue #274, opened 2019-05-10
Easy way to fix missing tool packages


matkoch:
If a tool package is missing, the assertion message currently says `Could not find package 'xxx' via 'yyy'`. We could introduce a new way to easily fix this, for instance via `nuke :fix` (global tool) or `nuke --fix` (`NukeBuild` base class), which would then automatically add the package reference to the build project file with the latest version.

Especially with packages that integrate with `build` assets or as `DotNetTool` type, we would also need to add proper excludes, for instance:

```
<PackageReference Include="coverlet.console" Version="1.5.1">
  <ExcludeAssets>all</ExcludeAssets>
</PackageReference>
```

PR #461, opened 2020-04-03
Add 'nuke :fix' command to add previously missing packages 

tunger:
Introduced ":fix" switch for nuke global tool, which lets you add a "PackageDownload" reference to your build project for previously missing tools in a failed build.

@matkoch Please let me know if this is what you had in mind.

There are some issues with which I'm not quite yet happy:
- way to determine .net core
- no good way to get latest version of a package
- I'm not sure how to detect if PackageDownload is supported. Does it depend on the dotnet runtime which executes the build, or the target framework of the build project? From what version on?

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common/Tooling/ToolPathResolver.cs
+++ b/source/Nuke.Common/Tooling/ToolPathResolver.cs
@@ -1,11 +1,13 @@
-﻿// Copyright 2019 Maintainers of NUKE.
+// Copyright 2020 Maintainers of NUKE.
 // Distributed under the MIT License.
 // https://github.com/nuke-build/nuke/blob/master/LICENSE
 
 using System;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using JetBrains.Annotations;
+using Nuke.Common.IO;
 using Nuke.Common.Utilities;
 using Nuke.Common.Utilities.Collections;
 
@@ -18,6 +20,8 @@ namespace Nuke.Common.Tooling
         public static string NuGetPackagesConfigFile;
         public static string NuGetAssetsConfigFile;
         public static string PaketPackagesConfigFile;
+
+        internal const string MissingPackageDefaultVersion = "latest";
 
         [CanBeNull]
         public static string TryGetEnvironmentExecutable(string environmentExecutable)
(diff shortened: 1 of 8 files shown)


matkoch:
- First check if `packages.config` exists. If it doesn't it's probably an SDK-based project.
- Then I guess the project needs to be parsed, to check what target framework is specified. I'm not sure if only `netcoreapp3.x` supports `PackageDownload`. That's to be investigated.
- `NuGetPackageResolver.GetLatestPackageVersion` can help to find the latest package version. I would just pick the latest release version.
- The exception type and changes in `BuildExecutor` are not as I pictured them. I would actually just modify the `GetPackageDirectory` method: generate the file, and include information in the exception message to call `nuke :fix`.
- I'd call the file `missing-package.log`

matkoch:
Have you already tried adding the change to the project? How does it look like after `project.Save()` ? Best case, it would sort it correctly to the list of already existing package references/downloads.

We probably also need to allow adding a package without a version. See https://github.com/RocketSurgeonsGuild/Nuke/blob/master/.build/.build.csproj

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/RocketSurgeonsGuild/Nuke">RocketSurgeonsGuild/Nuke</a></strong></div><div>Integration with the nuke build system. Contribute to RocketSurgeonsGuild/Nuke development by creating an account on GitHub.</div></blockquote>

tunger:
Included your comments.

It seems like the PackageDownload feature is dependent on the nuget version. https://github.com/NuGet/Home/wiki/%5BSpec%5D-PackageDownload-support
It was added in version 5, which was even added in dotnet sdk 2.1 (https://docs.microsoft.com/en-us/nuget/release-notes/nuget-5.0-rtm). Can it be?
However, I now let the user choose how he wants to add the package (reference/download).

Yep, the Microsoft.Build.Framework.Project correctly adds the item near existing tags. I was able to improve it to retain formatting and setting the metadata directly as attributes. So only the added line shows up as diff. Nice :)

Thanks for the hint with NuGet, it now queries for versions, and I let the user choose if he wants no version as well. Didn't know that was possible.

As for BuildExecutor, I thought it's more fitting there than writing files in the "GetPackageDirectory" utility function. I reverted the change.

<blockquote><div> GitHub</div><div><strong><a href="https://github.com/NuGet/Home">NuGet/Home</a></strong></div><div>Repo for NuGet Client issues. Contribute to NuGet/Home development by creating an account on GitHub.</div></blockquote>
<blockquote><div><strong><a href="https://docs.microsoft.com/en-us/nuget/release-notes/nuget-5.0-rtm">NuGet 5.0 RTM Release Notes</a></strong></div><div>Release notes for NuGet 5.0 including known issues, bug fixes, new features, and DCRs.</div></blockquote>

source/Nuke.Common/Tooling/ToolPathResolver.cs
```diff
@@ -67,7 +69,7 @@ string GetFramework(string file)
 
         private static string GetPackageDirectory(string[] packageIds, string version)
         {
-            return packageIds
```
matkoch, line 70: I would wrap the original statement with a try-catch... Then rethrow with the message modified.

source/Nuke.GlobalTool/Program.Fix.cs
```diff
@@ -0,0 +1,180 @@
...
+            if (IsProjectTypeValidForFix(buildProjectFilePath.Parent))
+            {
+                Logger.Error("Only .NET Core projects are supported.");
+                return 0;
+            }
+
+            var project = ProjectModelTasks.ParseProject(buildProjectFilePath).NotNull();
+            var packages = missingPackageContents
+                .Skip(1)
+                .Where(x => Regex.IsMatch(x, "^[^@]+@.+$"))
```
matkoch: I'm not sure what this is :)

source/Nuke.Common/Tooling/ToolPathResolver.cs
```diff
@@ -130,5 +144,21 @@ public static string GetPathExecutable(string pathExecutable)
                 .FirstOrDefault(File.Exists)
                 .NotNull($"Could not find '{pathExecutable}' via '{locateExecutable}'.");
         }
+
+        private static void WriteMissingPackageFile(IEnumerable<string> packageIds, string version)
+        {
+            var contents =
+                new[] { NukeBuild.BuildProjectDirectory.GlobFiles("*.csproj").FirstOrDefault()?.ToString() }
+                .Concat(packageIds.Select(x => $"{x}@{version ?? "latest"}"))
```
matkoch: Let's use `undefined` instead of `latest` here.

matkoch:
@david-driscoll is the props file for global package management always the same?

> We probably also need to allow adding a package without a version. See https://github.com/RocketSurgeonsGuild/Nuke/blob/master/.build/.build.csproj

source/Nuke.Common/Tooling/ToolPathResolver.cs
tunger: done

source/Nuke.GlobalTool/Program.Fix.cs
tunger: rewrote it to make clearer what's happening.

source/Nuke.Common/Tooling/ToolPathResolver.cs
tunger: done as suggested.

matkoch:
@tunger squashed all of your commits and force-pushed.

matkoch:
Related to https://github.com/microsoft/msbuild/issues/5263

matkoch:
Fixes #274

matkoch:

PR #464, opened 2020-04-05
Write out error on trapped error

batzen:
Currently trapped errors just cause the script to exit without any info about what failed where.
This PR adds code to write out the error which occurred.


matkoch:

PR #491, opened 2020-04-28
Issue 459

moerwald:

- [x] Is based on my own work

- I extracted GitUrl issues from GitRepository to GitUrlCreator, since it shouldn't be the responsibility of GitRepository to think about URL issues. 
- In GitVersionAttribute the code now writes a warning log entry in case of NoFetch == false && RepoUrl == SSH.
- Also added some comments in InjectValues.

Hope that's ok.


matkoch:
I'm sorry, but there is way too much code added which is unrelated. A PR should be as minimal as possible.

PR #500, opened 2020-05-02
Add ASCII support to OutputSink

Marat-Gumerov:

- [x] Is based on my own work

This pull request contains refactored code from #413 but uses other approach. ASCII support is implemented inside OutputSink, instead of inheritance. So inheritors could use ASCII mode. ASCII mode is turned on using environment variable.

It should help in resolving #412


matkoch:
Shouldn't be necessary with 0.25.0

PR #510, opened 2020-05-21
Case insensitive tool resolver

iwillspeak:

- [x] Is based on my own work


iwillspeak:
This won't build right now as the `EnumerationOptions` stuff is not in .NET Framework 4.7.2. I don't want to go around changing the target frameworks on `Nuke.Common` without more knowledge of _why_ it is set up the way it is.

matkoch:
I don't think it's a good idea to magically treat the file-system case-insensitive.

PR #518, opened 2020-05-30
Change Azure attribute to not use  nullable bool

lillo42:
Azure attribute is using  nullable bool and that is not allow in C#.

- [x] Is based on my own work


matkoch:
what is the issue here?

lillo42:
@matkoch .NET not allow bool? in attribute.

PR #525, opened 2020-06-19
Fix verbosity mapping for MSBuild/DotNet and add three new mappings

Yozer:

- [x] Is based on my own work

1.  It looks that there was a bug in the default verbosity mapping attribute for MSBuild and DotNet tasks.
2. I added three more mappings for NUnit, OpenCover and RaportGenerator,

### DIFF
--- a/source/Nuke.Common/Tools/ReportGenerator/ReportGeneratorTasks.cs
+++ b/source/Nuke.Common/Tools/ReportGenerator/ReportGeneratorTasks.cs
@@ -1,11 +1,25 @@
-// Copyright 2019 Maintainers of NUKE.
+﻿// Copyright 2019 Maintainers of NUKE.
 // Distributed under the MIT License.
 // https://github.com/nuke-build/nuke/blob/master/LICENSE
 
+using JetBrains.Annotations;
 using Nuke.Common.Tooling;
 
 namespace Nuke.Common.Tools.ReportGenerator
 {
+    [PublicAPI]
+    public class ReportGeneratorVerbosityMappingAttribute : VerbosityMappingAttribute
+    {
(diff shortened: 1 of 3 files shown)


matkoch:
As for `MSBuildTasks` and `DotNetTasks`... this is not a bug, this is the default of the two tools.

source/Nuke.Common/Tools/ReportGenerator/ReportGeneratorTasks.cs
```diff
@@ -1,11 +1,25 @@
...
 // Distributed under the MIT License.
 // https://github.com/nuke-build/nuke/blob/master/LICENSE
 
+using JetBrains.Annotations;
 using Nuke.Common.Tooling;
 
 namespace Nuke.Common.Tools.ReportGenerator
 {
+    [PublicAPI]
+    public class RaportGeneratorVerbosityMappingAttribute : VerbosityMappingAttribute
```
GeorgDangl: Shouldn’t it be ‚Report‘ instead of ‚Raport‘?

Yozer: Yes, you're correct. Nice catch.
- [ ] fix a typo

Yozer:
> As for `MSBuildTasks` and `DotNetTasks`... this is not a bug, this is the default of the two tools.

Ohh I didn't know that. Is this default because nuke wanted to have it in this way?
I think that default logging for msbuild is normal.

Anyway I can revert that, fix a typo so we will have few more mappers build-in.
What do you think about new mapping - especially:
```csharp
Minimal = nameof(ReportGeneratorVerbosity.Warning);
```

PR #535, opened 2020-07-14
Support importing Azure Pipelines system variables

mattbrailsford:
This PR adds support for importing any Azure Pipeline variables that aren't atuomatically mapped to environment variables (such as the system ones defined here https://docs.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml) as well as providing a way of mapping those variables. To import a variable you define a property and decorate it with a new `AzurePipelinesVariableAttribute`. This will look for azure pipelines variable with the same name as the property. If you'd like to map the property to a different pipeline variable, you can provide a variable name to the `AzurePipelinesVariableAttribute`'s constructor.

    [AzurePipelines(AzurePipelinesImage.Vs2017Win2016,
        InvokedTargets = new[] { nameof(Push) })]
    class Build : NukeBuild
    {
        [Parameter("The build number")]
        [AzurePipelinesVariable("Build.BuildNumber")]
        readonly string BuildNumber;

        [Parameter("The Azure Pipelines access token")]
        [AzurePipelinesVariable("System.AccessToken")]
        readonly string AzurePipelinesAccessToken;

        [Parameter("The GitHub API key")]
        [AzurePipelinesVariable]
        readonly string GitHubApiKey;
    }

When rendered, this then outputs as env variables like so

    ...
    - job: Compile
      displayName: 'Compile'
      dependsOn: [ Restore ]
      steps:
        - task: CmdLine@2
          inputs:
            script: './build.cmd Compile --skip'
          env:
            BuildNumber: $(Build.BuildNumber)
            AzurePipelinesAccessToken: $(System.AccessToken)
            GitHubApiKey: $(GitHubApiKey)
    ...

This PR also includes a minor fix for the VariableGroups as the rendered variable group should be wrapped in quotes.

Lastly, I've updated the unit test for this, however I don't seem to be able to run them on my local machine. The test runner just keeps skipping the test for some reason.

- [x] Is based on my own work


mattbrailsford:
PS This replaces the previous `ImportSecrets` parameter and the `ImportSystemAccessTokenAs` as these can both be accomplished with the one approach now.

mattbrailsford:
Hmmm, how would we feel about dropping the ImportVariables parameter on AzurePipelinesAttribute all together and just say any parameter that has the AzurePipelinesVariableAttribute on it is automatically mapped to be imported? It saves needing to declare the import twice.

matkoch:
As discussed privately, this is not exactly the direction we want to go. Thanks anyways!

PR #544, opened 2020-08-14
Fix target invocation for overridden targets

tunger:
When a build class overrides a target from an interface, executing the target via its name (command line) fails.

```csharp
        private interface ITestBuild
        {
            public Target Duplicate => _ => _
               .Executes(Action);
        }
        private class DuplicateTargetTestBuild : NukeBuild, ITestBuild
        {
            public Target Duplicate => _ => _
               .Executes(Action);
        }
```

Command `nuke duplicate` fails. This PR fixes it.

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common.Tests/Execution/DefaultInterfaceExecutionTest.cs
+++ b/source/Nuke.Common.Tests/Execution/DefaultInterfaceExecutionTest.cs
@@ -95,6 +95,17 @@ namespace Nuke.Common.Tests.Execution
             Assert.Throws<InvalidCastException>(() => ExecutableTargetFactory.CreateAll(build, x => x.E));
         }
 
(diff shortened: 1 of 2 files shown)


matkoch (commented):
test

matkoch:
@tunger the test doesn't cause an exception if I remove the `Distinct(x => x.Name)` ... I guess what happened is that you had the following case:

```
        private interface ITestBuild
        {
            public Target Duplicate => _ => _
               .Executes(Action);
        }
        private class DuplicateTargetTestBuild : NukeBuild, ITestBuild
        {
            // Note this is not public
            Target Duplicate => _ => _
               .Executes(Action);
        }
```

Could that be true?

PR #564, opened 2020-09-30
Support for custom exit codes #449

### LINKED ISSUE #449: Exit codes
### Description
It would be nice to provide ability to return custom exit codes. For example, in some case my global tool returns code 123 and then it would be easy to get this in jenkins pipeline script(def code = bat script: mytool SomeTarget, returnStatus: true) for next processing.
To do this now i found an ugly workaround:
image

maartenba:

- [x] Is based on my own work

Fixes #449. As discussed, setting an exit code to any `non-0` value will also influence `IsSuccesful`.

### DIFF
--- a/source/Nuke.Common/NukeBuild.cs
+++ b/source/Nuke.Common/NukeBuild.cs
@@ -70,13 +70,13 @@ namespace Nuke.Common
         /// When set to <value>null</value> (default), <see cref="Execute{T}"/> will return a <em>0</em> exit code on build success; or a <em>-1</em> exit code on build failure.
         /// When set to a non-null value, <see cref="Execute{T}"/> will return the value of <see cref="ExitCode"/>.
         /// </summary>
-        protected int? ExitCode { get; set; }
+        protected static int? ExitCode { get; set; }
         
         /// <summary>
         /// Executes the build. The provided expression defines the <em>default</em> target that is invoked,
         /// if no targets have been specified via command-line arguments.
         /// </summary>
-        protected int Execute<T>(params Expression<Func<T, Target>>[] defaultTargetExpressions)
+        protected static int Execute<T>(params Expression<Func<T, Target>>[] defaultTargetExpressions)
             where T : NukeBuild
         {
             var buildExitCode = BuildManager.Execute(defaultTargetExpressions);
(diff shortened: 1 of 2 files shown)


matkoch:
Are you a bot?

source/Nuke.Common/NukeBuild.cs
```diff
@@ -65,14 +65,23 @@ namespace Nuke.Common
     [SerializeBuildServerState]
     public abstract partial class NukeBuild : INukeBuild
     {
+        /// <summary>
+        /// Gets or sets the build exit code.
+        /// When set to <value>null</value> (default), <see cref="Execute{T}"/> will return a <em>0</em> exit code on build success; or a <em>-1</em> exit code on build failure.
+        /// When set to a non-null value, <see cref="Execute{T}"/> will return the value of <see cref="ExitCode"/>.
+        /// </summary>
+        protected static int? ExitCode { get; set; }
```
matkoch: Can you move this after `IsSuccessful` and make it `public` (non-static) ?

source/Nuke.Common/NukeBuild.cs
```diff
@@ -65,14 +65,23 @@ namespace Nuke.Common
...
         /// Executes the build. The provided expression defines the <em>default</em> target that is invoked,
         /// if no targets have been specified via command-line arguments.
         /// </summary>
         protected static int Execute<T>(params Expression<Func<T, Target>>[] defaultTargetExpressions)
             where T : NukeBuild
         {
-            return BuildManager.Execute(defaultTargetExpressions);
+            var buildExitCode = BuildManager.Execute(defaultTargetExpressions);
+            
+            return ExitCode ?? buildExitCode;
```
matkoch: When `ExitCode` is non-static, this logic should be moved to `BuildManager` .... This method here is just for delegation.

matkoch (changes requested):

source/Nuke.Common/NukeBuild.cs
maartenba: There you go

source/Nuke.Common/NukeBuild.cs
maartenba: Absolutely, done!

matkoch:
I like it :) Thanks 🤗

Issue #543, opened 2020-08-13
Please upgrade to the new Key Vault client libraries


heaths:
We have released [new versions](https://aka.ms/azsdk/net) of the Azure SDK Client Libraries for Key Vault, Storage, and other Azure services. We see that you are still using the older packages. The new Azure SDK is faster, leaner, and more extensible. Please see our [introductory video](https://aka.ms/azsdk/intro) and our [value propositions](https://aka.ms/azsdk/valueprop) for more information. We hope that you upgrade to these new packages. We are happy to offer assistance in upgrading while we work on migration guides.

For Key Vault, we have split the client libraries into 3 separate packages that can be used together but, individually, are smaller dependencies:

* [Azure.Security.KeyVault.Certificates](https://www.nuget.org/packages/Azure.Security.KeyVault.Certificates/) ([README](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/keyvault/Azure.Security.KeyVault.Certificates/README.md), [Samples](https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/keyvault/Azure.Security.KeyVault.Certificates/samples))
* [Azure.Security.KeyVault.Keys](https://www.nuget.org/packages/Azure.Security.KeyVault.Keys/) ([README](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/keyvault/Azure.Security.KeyVault.Keys/README.md), [Samples](https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/keyvault/Azure.Security.KeyVault.Keys/samples))
* [Azure.Security.KetVault.Secrets](https://www.nuget.org/packages/Azure.Security.KeyVault.Secrets/) ([README](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/keyvault/Azure.Security.KeyVault.Secrets/README.md), [Samples](https://github.com/Azure/azure-sdk-for-net/tree/master/sdk/keyvault/Azure.Security.KeyVault.Secrets/samples))

Samples are also available in our [samples browser](https://aka.ms/azsdk/net/keyvault/samples).

totollygeek:
I am also interested in this, since the old libraries pose a security problem, since they reference older NuGet packages, which include security vulnerabilities. Our security system is marking our repositories as vulnerable due to this. I will prepare a PR with the update.

GeorgDangl:
Maybe also consider including #703 then as well😀

totollygeek:
[...]

PR #575, opened 2020-10-16
Change KeyVaultTasks to use new SDKs

arodus:
Update `KeyVaultTasks` and related attributes to use the new Azure Key Vault SDK. This change breaks existing projects using the current implementation as the `TenantId` is now mandatory.

References: #543
 
- [x] Is based on my own work


GeorgDangl:
Looking forward to this😀

PR #579, opened 2020-10-18
Add EnvironmentInfo.IsDocker

arodus:
Add EnvironmentInfo to determine if Nuke is running in a (linux) docker container. Unfortunately, I'm not aware of a way to reliably detect if we're running in a windows container.

- [x] Is based on my own work

### DIFF
--- a/.github/workflows/continuous.yml
+++ b/.github/workflows/continuous.yml
@@ -30,6 +30,7 @@ jobs:
             GitHubToken: ${{ secrets.GITHUB_TOKEN }}
             SlackWebhook: ${{ secrets.SlackWebhook }}
             GitterAuthToken: ${{ secrets.GitterAuthToken }}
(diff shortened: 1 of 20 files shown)


Sam13:
@arodus In Windows container environment I check the existence of the service _cexecsvc_ to decide whether I'm inside a container or not.

Source: https://stackoverflow.com/questions/43002803/detect-if-process-executes-inside-a-windows-container

<blockquote><div>Stack Overflow</div><div><strong><a href="https://stackoverflow.com/questions/43002803/detect-if-process-executes-inside-a-windows-container">Detect if process executes inside a Windows Container</a></strong></div><div>It's simple. 
I would like to detect with code if my process is running inside a windows container. There are examples but they are all for linux based containers.

I'm looking for something unique...</div></blockquote>

PR #582, opened 2020-10-22
add fractional portion of second to build summary

zsd4yr:
This PR adds milliseconds to the build summary. Many of our targets complete in sub-second time and the current summary indicates that they took 0:00 time, which is a bit misleading imho

- [ X] Follows the contribution guidelines
- [ X] Is based on my own work
- [ X] Is in compliance with my employer


zsd4yr:
@matkoch I'm done with the PR and ready for review.

CI/CD completed except for the appveyor one. Not sure what is up with that...

zsd4yr:
@matkoch do you need anything else from me here? It is unclear to me whether or not I should try to resolve any issues in the continuous-integration/appveyor/pr build -- the travis-ci one appears to be passing

matkoch:
No it’s alright. I’m still unsure about the change though. When I published
one of the first versions of nuke, I remember someone saying “also it
doesn’t overwhelm you with a ridiculous precision for the duration of your
targets”... some other tools do this. I still get your point, but for many
other use cases, the precision is unnecessary.... maybe this should be more
reactive and only show higher precision when at least one target took less
than a second?!

zsd4yr:
@matkoch that makes sense, or perhaps the format of this summary should be an optional parameter when calling the nuke exe?

matkoch:
Mhhh I don't think so.... if anything, then it should work automatically, by checking all durations if one is less then 1 second

matkoch:
https://github.com/nuke-build/nuke/commit/0711430a00dcde72e2aad2b13112db10caa9f5dd

PR #594, opened 2020-10-28
Minor performance improvements

eduherminio:
* Remove unneeded `.ToList()` invocations.
* Avoid unnecessary zero-length array allocations by using `Array.Empty<T>()`.

---

- [✔️] Follows the contribution guidelines
- [✔️] Is based on my own work
- [✔️] Is in compliance with my employer

### DIFF
--- a/source/Nuke.MSBuildTasks/ExternalFilesTask.cs
+++ b/source/Nuke.MSBuildTasks/ExternalFilesTask.cs
@@ -53,7 +53,6 @@ namespace Nuke.MSBuildTasks
             return ExternalFiles
                 .SelectMany(GetExternalFilesInfo)
                 .Select(DownloadExternalFile)
-                .ToList()
                 .All(x => x.Result);
         }
 
--- a/build/Build.Contributors.cs
+++ b/build/Build.Contributors.cs
@@ -22,7 +22,7 @@ partial class Build
     Target UpdateContributors => _ => _
         .Executes(() =>
         {
-            var previousContributors = File.Exists(ContributorsCacheFile) ? ReadAllLines(ContributorsCacheFile) : new string[0];
+            var previousContributors = File.Exists(ContributorsCacheFile) ? ReadAllLines(ContributorsCacheFile) : System.Array.Empty<string>();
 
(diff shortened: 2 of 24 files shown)


source/Nuke.MSBuildTasks/ExternalFilesTask.cs
```diff
@@ -53,7 +53,6 @@ protected override bool ExecuteInner()
             return ExternalFiles
                 .SelectMany(GetExternalFilesInfo)
                 .Select(DownloadExternalFile)
-                .ToList()
```
matkoch, line 56: I'm pretty sure this is incorrect here.

matkoch:
And tbh, I rather have the array allocations than writing a tedious `Array.Empty<T>()` every time. None of the code is really performance-critical, so I'd go with the better-looking way.

source/Nuke.MSBuildTasks/ExternalFilesTask.cs
eduherminio: hmm I'm surely missing something 😅 
Can I ask why? Does it affect the `.All()` in the next line?

matkoch: `DownloadExternalFile` returns a `Task<T>` ... that `.ToList()` is required to trigger the parallelization, otherwise it will just sequentially check the returned results.

eduherminio:
> And tbh, I rather have the array allocations than writing a tedious `Array.Empty<T>()` every time. None of the code is really performance-critical, so I'd go with the better-looking way.

Sure 😄 closing it.

matkoch:
Thanks anyways!

source/Nuke.MSBuildTasks/ExternalFilesTask.cs
eduherminio: That makes sense, thanks for the clarification.

Nefarion: But there is no AsParallel() to start parallelization? Or am i missing something?

PR #629, opened 2020-12-16
Add DocFXTasks custom logger

billhenn:
The changes add a custom logger for DocFX to properly report warning and error messages.

- [x] Is based on my own work


bitbonk:
I still don't know enough about NUKE, but in regards to  docfx this looks great to me. When using Nuke this will solve https://github.com/dotnet/docfx/issues/6894

matkoch:
@billhenn thanks for the PR, I have concerns though regarding the regex :)

1. It's not clear to me if with a different culture, this regex will actually fail.
2. Given that the regex has a fixed number of digits, it seems easier to just use `Substring` guarded with `str.Length`.

If we use the `Substring` approach, please check how the time is printed. If it's using the `CultureInfo`, we might just check and save the length to proceed as described.

billhenn:
@matkoch I don't know DocFX's codebase at all but their console writing logic seems to be in the Microsoft.DocAsCode.Common.ConsoleLogListener class and is implemented via this line:

```
var now = DateTime.UtcNow.ToString("yy-MM-dd hh:mm:ss.fff");
```

It's your call how you want to handle the regex.  I used it because I could easily detect the various warning/error and that all this occurred at the start of a line.

matkoch:
Regex matching takes much more time than using substring. It's just unnecessary here imo.

PR #633, opened 2020-12-21
Add more Space Automation support

maartenba:
Adds more Space Automation support for basic scenarios.

- [x] Is based on my own work


matkoch:
Thanks @maartenba! I took this as a reference and adjusted it where necessary.

One thing that got me curious is why `volumeSize` is on the `container` level, while it sounds like it should be available under `resources`. Is this a mistake or really intended on Space Automation  side?

PR #676, opened 2021-02-08
CorFlags CLI tool

chaquotay:
CorFlags CLI tool integration.

I named the 32bit flags according to [` System.Reflection.PortableExecutable.CorFlags`](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.portableexecutable.corflags?view=net-5.0), since their "natural" name (e.g. `32BitPref`) doesn't work in NUKE, because the name is not valid in the generated C# code.

- [x] Is based on my own work


matkoch:

PR #700, opened 2021-03-08
Updated custom logger to handle errors and warnings from both angular and npm

alex-jones-rsk:
Summary:

When building and testing angular projects with nuke we found that information logs were being output as errors. This would fail our builds on Azure DevOps. I'd like to suggest this change to the CustomLogger for NPM to be more generalized when catching errors, warnings and normal logs.

- [x] Is based on my own work

### DIFF
--- a/build/Build.Enterprise.Enable.cs
+++ b/build/Build.Enterprise.Enable.cs
@@ -21,9 +21,9 @@ partial class Build
 {
     [Parameter] readonly string EnterpriseAccessToken;
 
-    public class RestartWithEnterpriseAttribute : BuildExtensionAttributeBase, IOnBeforeLogo
+    public class RestartWithEnterpriseAttribute : BuildExtensionAttributeBase, IOnBuildCreated
     {
-        public void OnBeforeLogo(NukeBuild build, IReadOnlyCollection<ExecutableTarget> executableTargets)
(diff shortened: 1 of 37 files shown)


matkoch:
Thanks for the PR. Doesn't look much performant though to check the full lines for appearance of strings. Also what happens if a file or ordinary string contains "warn"? I'm sure that should not result in a warning.

GeorgDangl:
There was this issue present on the Angular repository:
https://github.com/angular/angular-cli/issues/16994
Basically, `stderr` doesn't necessarily mean an error is logged. But that's confusing😀

I don't think that's the right approach here. @matkoch is right, any string containing `warn` would be logged as error.

Imho, the right approach would be to fix it upstream at the tool, and as a workaround, just use something like this in the build:

```csharp
    public Build()
    {
        // Otherwise, Angular compiler output is logged to stderr instead of stdout.
        NpmTasks.NpmLogger = (outputType, message) => Logger.Normal(message);
    }
```

@matkoch, what do you think, should we maybe put that info in the docs?

matkoch:
@GeorgDangl I would rather like to get PR to fix these issues to allow better OOTB behavior.

@alex-jones-rsk I'm closing this for now, but if you can suggest a more accurate detection, feel free to open another PR.

PR #711, opened 2021-03-14
Add DotNetBuild as a copy of MSBuild

andrewlock:
I don't really understand how the codegen works here, so was just trying something out... The source generation only seems to work occasionally, so not sure if I'm doing something wrong. I tried running `./build.ps1 GenerateTools` but that seems to update lots of files, and always errors with `Couldn't load metadata file 'ILRepack.json'`

As already discussed, not sure if this is what you want, or if there's a better way of duplicating MSBuild for `dotnet msbuild`, but this is going to be a blocker for me, so wanted to try something out! 🙂

I confirm that the pull-request:

- [ ] Follows the contribution guidelines
- [ ] Is based on my own work
- [ ] Is in compliance with my employer

PR #713, opened 2021-03-14
Alternative approach to adding dotnet msbuild

andrewlock:
This one avoids the duplication, but puts code that's usually autogenerated in `DotNetTasks`. Not sure if this will one will work either?

- [ ] Follows the contribution guidelines
- [ ] Is based on my own work
- [ ] Is in compliance with my employer


matkoch:
See https://github.com/nuke-build/nuke/pull/711#issuecomment-799903927

PR #717, opened 2021-03-17
Allow querying MSBuild properties of legacy web projects

macsux:
Fixes issue where old-style ASP.NET web projects cannot be processed via `ProjectModelTasks.ParseProject` as it requires extra targets that are not part of .NET core SDK.


matkoch:
Think I don't want to pull any more legacy support into the project.

PR #723, opened 2021-03-25
Allows creating properties that act as members on INukeComponent, simply value injection syntax

macsux:
Allows INukeBuild components to declare stateful field-like members, and simplify syntax for value injectors in components
Example:
```csharp
interface IMyNukeComponent : INukeBuild
{
    [GitVersion] GitVersion Versioning => this.Get(); // Value injected readonly member
    [Parameter] string MyParameter1 => this.Get(); // Parameter 
    [Parameter] string MyParameter2 => this.Get("value"); // Parameter with default value
    
    // equivalent of a field on concrete Build class that can be used to pass state across targets
    string StatefulMember 
    { 
        get => this.Get(); 
        set => this.Set(value); 
    } 
}
```

- [x ] Follows the contribution guidelines


matkoch:
Relying on `Stacktrace` isn't a good idea because a property might get inlined. `CallerMemberName` is the way to go.

`Get("value")` as a way to define a fallback value is ambiguous in what it means. if anything it should be `GetValue() ?? "default"`.

matkoch:
Anyways, this hack of saving the `Expression<>` won't work. Different interfaces can have identically named properties which get different values injected. In the extension method there is no way to distinguish from which type the call came.

Basically, this is missing: https://github.com/dotnet/csharplang/discussions/87

macsux:
Crap, I forgot about inlining, but you're right. What about an alternate approach. Put this logic into a code generator, as we already introduced one. It would look for all properties declared on interfaces that inherit from `INukeBuild` without a body that has any of the `ValueInjectionAttributeBase` attribute applied and automatically synthesize an explicit interface implementation for the property on the class like and prevent declaring collision issue you brought up
```csharp
int IMyComponent.MyProperty 
{ 
    get => Get("IMyComponent.MyProperty");
    set => Set("IMyComponent.MyProperty",value);
}
```
That should solve this nicely, and actually be even more elegant than what this PR does. The only downside is providing defaults, but if we add the idea of `INukeBuildEventsAware` that would solve this, as components can just put all defaults into one of the early event hooks to provide defaults.

matkoch:
This approach has way too many implications. Off-hand: 1) build must be marked partial 2) existing overrides must be considered. Plus I don't see the status quo being unbearable complicated. You can use snippets (available for Rider and ReSharper), or just static fields...

PR #743, opened 2021-04-15
Allow to set description from derived attributes

cmenzi:
When executing `--help` to get the parameter descriptions, all KeyVaulSecrets had no description, because it was not possible to set a `Description`. As the `ParameterAttribute` is intended to be derived and it anyway allows to set null as description, we may can allow to use the setter.

This should not break existing implementations. The other option would be, that all derived ParamaterAttributes should have proper c'tors with a description and call the correct base c'tor.

- [x] Is based on my own work


matkoch:
did you read the updated pr template?

matkoch:
as discussed

PR #761, opened 2021-06-30
Fix version links in CHANGELOG.md

glconti:

- [x] Follows the contribution guidelines (not sure, since I am not a sponsor, nor a blogger, nor greatly involved in any OOS project, just a Nuke user and fan. If you reject the PR I'll understand, no problem)
- [X] Is based on my own work

I was browsing the Nuke CHANGELOG.md and all the links to the versions are broken, it seems there's one `.git` extra in each link, for example https://github.com/nuke-build/nuke.git/compare/5.2.0...5.2.1 results in a 404 for me while https://github.com/nuke-build/nuke/compare/5.2.0...5.2.1 works.


GeorgDangl (commented):
Looks good to me, but where is the error coming from? Are these links automatically generated? If yes, then we maybe also should update the way it's generated.

glconti:
> Looks good to me, but where is the error coming from? Are these links automatically generated? If yes, then we maybe also should update the way it's generated.

@GeorgDangl Good catch, I had already looked in the code and thought that just the latest version was generated, but after a second look, actually all the links are regenerated when the changelog is finalized in https://github.com/nuke-build/nuke/blob/6eefa9bd22bb1fcab9c8f9697a3db74019480a49/source/Nuke.Common/ChangeLog/ChangeLogTasks.cs#L225-L241

The issue here is that the property `repository.HttpsUrl` is made for the git protocol, not the browser https://github.com/nuke-build/nuke/blob/6eefa9bd22bb1fcab9c8f9697a3db74019480a49/source/Nuke.Common/Git/GitRepository.cs#L237-L238

What would you advise here? have a second property or trim the `.git` for this specific case?
I understand that's a low priority thing, so you tell me if you want to just create an issue for later or I could still bring this small fix.

GeorgDangl:
Imho we should just directly trim the extension for this use case. I don’t think having GitHub specific urls in the general case is something that we should do. 
but, I also think that there is some class, somewhere which can check if a repository is a GitHub repository and generate the url. 

But still, just trimming the extensions should be fine here😀

GeorgDangl:
And maybe with a small comment in the code where we trim the extension explaining why we do it, to generate GitHub links.

glconti:
Cool, I plan to implement it shortly

matkoch:
Fixed in https://github.com/nuke-build/nuke/commit/a8dd618d2dc35a1d311e06dd775d35f390053bc9. Thanks for bringing to attention.

PR #766, opened 2021-07-05
Add support for OctoVersion

matt-richardson:
This PR adds support for [OctoVersion](https://github.com/OctopusDeploy/OctoVersion) as a built in tool.

This was (briefly) discussed [back in Feb](https://nukebuildnet.slack.com/archives/C9Q99MFU0/p1614107457035200?thread_ts=1614060572.028500&cid=C9Q99MFU0).

The `Nuke.OctoVersion` package (that never got published to nuget) has been removed from OctoVersion (see https://github.com/OctopusDeploy/OctoVersion/pull/38) and this PR replaces it with first class support in Nuke.

I fully expect that there'll be some changes to be done - let me know what you think!

Thanks for reviewing :tada:

- [x] Is based on my own work

### DIFF
--- a/build/specifications/OctoVersion.json
+++ b/build/specifications/OctoVersion.json
@@ -1,0 +1,158 @@
+{
+  "$schema": "https://raw.githubusercontent.com/nuke-build/nuke/master/source/Nuke.CodeGeneration/schema.json",
+  "references": [
+    "https://github.com/OctopusDeploy/OctoVersion/blob/main/README.md"
+  ],
+  "name": "OctoVersion",
+  "officialUrl": "https://github.com/OctopusDeploy/OctoVersion",
+  "packageId": "OctoVersion.Tool",
+  "packageExecutable": "OctoVersion.Tool.dll",
+  "customExecutable": true,
+  "tasks": [
+    {
+      "help": "Gets the version information for a project.",
+      "postfix": "GetVersion",
+      "returnType": "OctoVersionInfo",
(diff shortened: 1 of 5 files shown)


matt-richardson:
From the Octopus side, @corey-underdown is also keeping an eye on this - just in case I'm not available.

matkoch:
Since you're also publishing for `netcoreapp3.1` I tend to suggest setting this as a default for `OctoVersionAttribute.Framework` because the current `GitVersionAttribute` also does that and therefore both would target LTS. Not that it means a lot to me, just a thought.

matt-richardson:
> Since you're also publishing for `netcoreapp3.1` I tend to suggest setting this as a default for `OctoVersionAttribute.Framework` because the current `GitVersionAttribute` also does that and therefore both would target LTS. 

I just went to do this, and noted that `GitVersionAttribute` [actually targets](https://github.com/nuke-build/nuke/blob/develop/source/Nuke.Common/Tools/GitVersion/GitVersionAttribute.cs#L26) `netcoreapp3.0`...

Even more interestingly, the [latest gitversion.tool](https://www.nuget.org/packages/GitVersion.Tool/5.6.11) doesn't actually support `netcoreapp3.0`, only `netcoreapp3.1` and `net5.0`.

As nuke is now net5, I'm thinking the better path forward here is to bump the gitversion framework instead. Thoughts?

matkoch:
agreed, i will bump gitversion

PR #773, opened 2021-08-01
Adds LFS and Submodule support to Azure Pipelines YAML

ChaseFlorell:

- [x] Is based on my own work

In this Pull Request, I've added support for both git LFS (Large File Storage) and git Submodules for Azure Pipelines. The details around azure pipelines lfs and submodule support in yaml can be [found in their documentation](https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/pipeline-options-for-git?view=azure-devops&tabs=yaml#checkout-submodules)


ChaseFlorell:
hmm, tests pass locally, not sure what I did wrong for them to fail CI
image

matkoch:
Hey Chase! Looks good except that the setters on the attribute should follow the same pattern as `TriggerBatch`. Only when we have a nullable bool, we can make sure that the requested configuration is also written to file. I can also address by myself, just let me know then.

ChaseFlorell:
Thanks @matkoch, I think I've got it right.

PR #778, opened 2021-08-13
Update OctoVersion to support passing NUKE parameters

matt-richardson:
When we integrated [OctoVersion](https://github.com/nuke-build/nuke/pull/766) support into nuke, we changed it from [being in-process](https://github.com/OctopusDeploy/OctoVersion/pull/38) to out of process. This meant that we were now using stdout to read all response, causing issues where json output was being intermingled with service messages to TeamCity.

Consequently we made some adjustments to the output formatters that support this, and [added](https://github.com/OctopusDeploy/OctoVersion/pull/47) support for passing a `--detectenvironment` argument, which is used to avoid hardcoding a CI output format.

This PR adjusts the `[OctoVersion]` attribute to:
* support the new approach and get parseable json output for emitting the `OctoVersionInfo` c# class from the tool execution, as well as auto detected output for the CI platforms supported
* support additional OctoVersion options specified as NUKE parameters so they can be configured to be passed at runtime by the calling of the Nuke build (as well as specified via static Attribute properties or left blank to allow OctoVersion's own overrides such as environment variables or config files, to function)  `Branch` `FullSemVer` `Major` `Minor` `Patch` are now supported.
* have smoother logic around providing an `AutoDetectBranch` option.  By design OctoVersion does not aim to do this and it must be told the branch, however the NUKE attribute approach will allow this to be calculated if the caller desires.  This can be useful to default to true when `NukeBuild.IsLocalBuild` to enable a good F5 experience for developers running the Nuke build locally.

- [x] Is based on my own work


ryangribble:
Ive pushed up a number of changes after chatting with @matkoch and @matt-richardson and others

- Added support for additional OctoVersion parameters so we can remove the need to set env vars before calling Nuke+OctoVersion.  `Branch` `Major` `Minor` `Patch` and `AutoDetectBranch` are now supported
- Tidied up support for the Nuke attribute automatically calculating the branch from local git folder when enabled
- added logging

### Additional Parameters
Working with @matkoch we also implemented each parameter both as a hardocded property on the attribute 
eg 
```
[OctoVersion(Branch = "my-feature")]
```

as well as a NUKE Property based approach, allowing the calling NUKE build script to define properties that can be passed to the attribute (thus allowing them to be specified at run time)

eg 
```
[Parameter]string OctoVersionBranch;
[OctoVersion(BranchParameter = nameof(OctoVersionBranch)]
```

Validation is carried out to ensure that the attributes are mutually eclusive.  eg if you specify `Branch` you cant specify `BranchAttribute` and vice versa.

Note that there was some shennanigans to implement the Major/Minor/Patch properties as they are `int?` (and similarly the AutoDetectBranchName being `bool?`) but named attribute properties cannot be nullable.   The properties need to be nullable because we want to be able to define them to MAYBE be passed in but if not specified then we simply dont set those arguments on the tool execution. The solution used here is to have a the public attribute property declared as `int` with get/set updating a private backing variable `int? _patch` which is working well.  We get compile time type checking on the NUKE parameter and attributes and run time checking of any incorrect values passed via paramerters.  eg the NUKE parameter has to be an `int?`, it can be passed to `[OctoVersion(PatchParameter = xxxx)]` OK, and specifying `--OctoVersionPatch "abc123"` results in an error as it can't be parsed to an int

### Calculating git branch
OctoVersion requires a branch to be provided and does no detection itself.  It can be specified via `Branch` or `BranchAttribute` or was being auto detected in the NUKE attribute.  The way this is done has been tidied up as per @matkoch suggestion as well as the logic/checking adjusted so that if you specify a branch and autodetect is true, an error is thrown.  We dont want incorrect versions to be calculated so want to throw errors when invalid combinations are detected

### Logging
Also added logging to the attribute as the logic occuring during the parameter injection is important to capture in build logs.  This output is also wrapped in a log block/section.  Trace logging is output for the definition of the various attribute parameters and thenInfo logging used for the actual resolved values:
image



Example github actions pipeline using these new features:

https://github.com/ryangribble/actions-tests/runs/3461957305?check_suite_focus=true

Build.cs definition sets up POSSIBLE parameters that may be set by the caller
image

GitHub Actions workflow yml snippet - we are only passing in Branch and Patch and leaving the others undefined (so they aren't passed to OctoVersion and it does it's own thing)
image

GitHub Actions log output (including verbose log level)
image


Meanwhile if we want to run locally we could use

`dotnet run -- --OctoVersionAutoDetectBranch`

image

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
```diff
@@ -22,27 +24,232 @@ namespace Nuke.Common.Tools.OctoVersion
...
-                .SetOutputFormats(OctoVersionOutputFormatter.Json)
-                .SetFramework(Framework)
-                .When(NukeBuild.IsLocalBuild, s => s
-                    .SetCurrentBranch(GitTasks.GitCurrentBranch())))
+                    .SetFramework(Framework)
+                    .SetOutputJsonFile(tempOutputFile)
+                    .When(UpdateBuildNumber, x => x.EnableDetectEnvironment())
+                    .When(!UpdateBuildNumber, x => x.SetOutputFormats("JsonFile"))
+                    .When(!string.IsNullOrEmpty(octoversionBranch), x => x.SetCurrentBranch(octoversionBranch))
+                    .When(octoVersionMajor != null, x => x.SetMajor(octoVersionMajor))
```
matkoch: `When` is unnecessary here

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
```diff
@@ -22,27 +24,232 @@ namespace Nuke.Common.Tools.OctoVersion
...
+            // Wrap log output in a NUKE Section
+            using (Logger.Block("[OctoVersion] Injection"))
+            {
+                LogAttributeProperties();
+
+                return GetValueInternal(member, instance);
+            }
+        }
+
+        private void LogAttributeProperties()
```
matkoch: Not a fan of this for three reasons:
1. We don't have log output like this anywhere else
2. There is an `AuditBuildMembers` attribute in `Nuke.Enterprise` which logs all the members, but it could be extended to also log the configuration of such attributes.
3. The `Logger.Block` is – in a sense – reserved for targets. Probably there should be two methods, like `Logger.Block` and `Logger.SubBlock`. Will keep a note.

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
```diff
@@ -22,27 +24,232 @@ namespace Nuke.Common.Tools.OctoVersion
...
+                octoVersionPatch = _patch;
+            }
+            else if (!string.IsNullOrEmpty(PatchParameter))
+            {
+                ControlFlow.Assert(_patch == null, $"If {nameof(PatchParameter)} is specified, then {nameof(Patch)} should not be specified.");
+                octoVersionPatch = GetParameterValueFromBuild<int?>(PatchParameter, instance);
+            }
+
+            // Log actual resolved values of parameters
+            Logger.Info("Resolved OctoVersion Properties");
```
matkoch: Same as with `LogAttributeProperties`

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
```diff
@@ -22,27 +24,232 @@ namespace Nuke.Common.Tools.OctoVersion
...
+            Logger.Trace($"  BranchParameter:           {BranchParameter ?? "null"}");
+            Logger.Trace($"  MajorParameter:            {MajorParameter ?? "null"}");
+            Logger.Trace($"  MinorParameter:            {MinorParameter ?? "null"}");
+            Logger.Trace($"  PatchParameter:            {PatchParameter ?? "null"}");
+        }
+
+        private object GetValueInternal(MemberInfo member, object instance)
+        {
+            // Get AutoDetectBranch / AutoDetectBranchParameter
+            bool? octoVersionAutoDetectBranch = null;
```
matkoch: These blocks here are quite repetitive. Also, I would consider the following:

For instance with `AutoDetectBranch`, a user might want to set a certain value, but we could as well allow the parameter `AutoDetectBranchParameter` to override it. So the logic is essentially `GetValue(AutoDetectBranchParameter) ?? AutoDetectBranch`, which I think is much more flexible.

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
```diff
@@ -22,27 +24,232 @@ namespace Nuke.Common.Tools.OctoVersion
...
+            }
+
+            if (string.IsNullOrEmpty(octoversionBranch))
+            {
+                // If we have ended up with no branch, abort!
+                ControlFlow.Fail($"OctoVersion requires the branch to be specified.  Either enable auto detection via {nameof(AutoDetectBranch)} or {nameof(AutoDetectBranchParameter)}, or one of {nameof(Branch)} or {nameof(BranchParameter)} must be specified.");
+            }
+            
+            // Now it's GO TIME
+            Logger.Info($"Executing OctoVersion.Tool.");
```
matkoch: Execution is logged anyways

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
```diff
@@ -22,27 +24,232 @@ namespace Nuke.Common.Tools.OctoVersion
...
                 AzurePipelines.Instance?.UpdateBuildNumber(version.FullSemVer);
+                //this also happens inside OctoVersion when running in TeamCity and EnableDetectEnvironment was set (based on UpdateBuildNumber)
                 TeamCity.Instance?.SetBuildNumber(version.FullSemVer);
                 AppVeyor.Instance?.UpdateBuildVersion(version.FullSemVer);
             }
 
             return version;
         }
+
+        private static T GetParameterValueFromBuild<T>(string parameterName, object target)
```
matkoch, line 161: This method I will implement in the base type, asap

source/Nuke.Common/Tools/OctoVersion/OctoVersionTasks.cs
```diff
@@ -31,16 +40,18 @@ internal static string GetToolPath(string framework = null)
 
         private static OctoVersionInfo GetResult(IProcess process, OctoVersionGetVersionSettings toolSettings)
         {
+            ControlFlow.Assert(!string.IsNullOrEmpty(toolSettings.OutputJsonFile), $"{nameof(toolSettings.OutputJsonFile)} must be set");
```
matkoch, line 39: So just for clarification, the output file is now mandatory? I can't invoke OctoVersion and just let it set the environment variables?

source/Nuke.Common/Tools/OctoVersion/OctoVersionTasks.cs
```diff
@@ -31,16 +40,18 @@ internal static string GetToolPath(string framework = null)
...
+
+            var output = "";
             try
             {
-                var output = process.Output.EnsureOnlyStd().Select(x => x.Text).ToList();
+                output = File.ReadAllText(toolSettings.OutputJsonFile);
```
matkoch: `SerializationTasks.JsonDeserializeFromFile<JObject>` ... afaik, it does no exception handling, but it's on my list. Anyway, I don't expect any issues with the output file, so better to use the central API

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
```diff
@@ -22,27 +24,232 @@ namespace Nuke.Common.Tools.OctoVersion
     [UsedImplicitly(ImplicitUseKindFlags.Default)]
     public class OctoVersionAttribute : ValueInjectionAttributeBase
     {
+        /// <summary>
```
matkoch, line 33: Member layout should be first all fields, and then the attributes

build/specifications/OctoVersion.json
```diff
@@ -13,79 +13,104 @@
...
+      "definiteArgument": "octoversion",
+      "settingsClass": {}
     }
   ],
+  "commonTaskPropertySets": {
+    "octoVersionSettings": [
+      {
+        "name": "CurrentBranch",
+        "type": "string",
+        "format": "--CurrentBranch {value}",
```
matkoch, line 37: Question for all the `format` properties, does OctoVersion also allow lisp-cased parameters, e.g., `--current-branch` ?

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
```diff
@@ -22,27 +24,232 @@ namespace Nuke.Common.Tools.OctoVersion
...
+        /// Which framework to use when selecting the OctoVersion library from the package
+        /// </summary>
         public string Framework { get; set; } = "net5.0";
-        public bool UpdateBuildNumber { get; set; }
+
+        /// <summary>
+        /// Whether to update the build number and output octoversion attributes into the CI platform
+        /// Supports AzurePipelines, TeamCity, AppVeyor and GitHubActions
+        /// </summary>
+        public bool UpdateBuildNumber { get; set; } = NukeBuild.IsServerBuild;
```
matkoch: This is dead-code I believe, because inside the if-branch it's using the null-shorthand like `<Host>.Instance?...`.

matkoch (changes requested):
Left individual comments

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: That implies a certain level of understanding of how the auto generated tool classes work.  
For anyone looking at the OctoVersion support and looking at this code here, this feels clearer that we dont pass these arguments to the tool when no value was resolved in the attribute.

If you strongly object I can remove the `.When` but I feel it does add clarity at the point where the tool is executed 🤣

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: Keep in mind it is trace level logging so only shows up when running with verbosity enabled.  It was useful to me when troubleshooting/consuming the attribute to understand what was being passed in.  The consumers of this will be looking at a build log in a CI tool like teamcity or github actions etc so it seemed like any information we can output as to what is going on would be helpful

The reason for attempting to put it in a LogBlock section is, that the attribute injection occurs BEFORE any build targets and having all of the logged output from the setting up to run the tool AND the actual tooll execution itself, floating on it's own before any sectioned build targets seemed a bit off putting.  Putting it in a log block means that it is expanded/collapsed in the supported CI platforms and much more clearly delineated that it is all of the relevant log output from the [OctoVersion] attribute injection.  
It also defaults to collapsed in GHActions...
so we end up with this:
image

the alternative would be we have quite a bit of log output (even if we remove the parameter logging i added here, just from the tool execution) that isnt collapsed and seems more cluttered.

if you implemented support for SubBlocks im not sure it would look much different here, as we arent in a block yet anyway (we are in the attribute injection phase, before any build targets are running).

I felt this was cleaner but if you want it to be removed then we can do that

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: Yes they are repetitive 🤣 . For the intial review i think it was a good idea to keep it all spelled out so we could discuss.
I like your idea of the override behaviour - not only does it make the code clearer it also removes alot of the ControlFlow assertions that were trying to make the pairs of properties mutually exclusive.  I will look to update it

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: 👍

source/Nuke.Common/Tools/OctoVersion/OctoVersionTasks.cs
ryangribble: Yes.  
When moving from our inhouse attribute to this new built in approach, the execution method was also changed from executing OctoVersion as a c# library to executing it as an autogenerated Nuke CLI tool.  Executing it on CLI means we need to parse the output as Json to get the c# object exposed to the rest of the c# Nuke build script (eg the `OctoVersionInfo versionInfo`).  The initial implementation in this PR was running OctoVersion twice (once with Json output for parsing, then a 2nd time with auto detected CI output to set variables in the CI platform).  This new implementation runs it once, using an output file for the parseable json output and the console output for the autodetected CI platform.
The file is created in the NUKE temp area and is deleted after it's parsed.

source/Nuke.Common/Tools/OctoVersion/OctoVersionTasks.cs
ryangribble: :+1:

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: By dead code are you referring to having the property at all?  Or having the default value set to `NukeBuild.IsServerBuild` ?

FYI we are using this parameter to set whether to autodetect the CI platform (which causes OctoVersion to emit variables into the CI platform... unrelated to Nuke specifically)... and also using the parameter in the Nuke attribute to specifically SetBuildNumber on the 3 CI platforms Nuke supports

build/specifications/OctoVersion.json
ryangribble: No, the parameters are case insensitive but they cant contain additional hyphens/underscores

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: implemented as per your suggestion :+1:

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: Ive removed the parameter logging but i really would like to keep the log block as it contains the output from the tool nicely! otherwise these value injected attributes are just floating off in no man's land

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: (i also had a look in the generated tool class and cant even see that it doesnt set the cmdline parameter when the passed value is null... so it must be even deeper in the execution flow if this does happen.  It feels alot clearer to me to leave these conditions at this obvious point where it sets up the tool execution

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: removed this line

source/Nuke.Common/Tools/OctoVersion/OctoVersionTasks.cs
ryangribble: I looked into doing this but the central API can't handle the deserialization as we need to use a `AllWritableContractResolver`.  Our approach here is the same as what Nerdbank and GitVersion are already doing.

I propose we leave this as is, and if any future work occurs to enhance the central API to handle this case, all 3 version tools can be updated to use the central API?

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
matkoch: Sorry, I just meant the default value. It can just be `true` because anyway on local execution there is nothing to execute it on. Maybe though we could set `Console.Title`, but that's a different topic.

source/Nuke.Common/Tools/OctoVersion/OctoVersionTasks.cs
matkoch: True, but maybe we can adapt that and add the resolver per default there too.

Alternatively – although not so clean – it could also be this:
```
SerializationTasks.JsonDeserialize(input, settings =>
{
    settings.ContractResolver = new AllWritableContractResolver();
    return settings;
});
```

Actually I wish we had a general `With` method.

source/Nuke.Common/Tools/OctoVersion/OctoVersionTasks.cs
matkoch: Understood. Did you also consider putting them all in env vars? Maybe even the whole json string? I'm just genuinely curious, because I also have a couple of files that are put into temp directory. So no objects to keep it as is.

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
matkoch: I understand the argument, but logging is a part that I'd like to come back to a bit later. In the meantime, I don't want to introduce any tracing or logging that could be solved in a more general and configurable way, as mentioned with the `AuditBuildMembers` attribute.

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
matkoch: Maybe I should communicate that better in the docs, but `SetX(null)` is essentially a `ResetX()`. It's probably important to understand that because it's supposed to avoid exactly this type of "excessive/repetitive/error-prone conditions".

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: that parameter is also being used to tell octoversion tool whether to auto detect environment or not.  However due to some other reasons ive decided to default to just `true` anyhow 👍

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: Done

source/Nuke.Common/Tools/OctoVersion/OctoVersionTasks.cs
ryangribble: ive pushed an update that enhances the Nuke `SerializationTasks.JsonDeserializeFromFile()` to take a `Configurator` and used that

source/Nuke.Common/Tools/OctoVersion/OctoVersionTasks.cs
ryangribble: we did consider a few options but this was the best option given the way octoversion works, how the old (internal) Nuke.OctoVersion worked and not wanting to change too many things 🤣 We will stick with the file for now

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: Removed

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: LogBlock now removed.  It would definitely be nice to see Nuke core do something here so that the attribute injection stuff is a little more clear and bounded, in the log outputs.  It's not a blocker for this PR though, so ive removed the LogBlock

source/Nuke.Common/Tools/OctoVersion/OctoVersionAttribute.cs
ryangribble: Ok I have removed all of those When clauses

ryangribble:
@matkoch i think all feedback should be addressed, let me know if you see anything further :+1:

PR #807, opened 2021-10-12
Add Unity version auto-detection

razenpok:
Unity projects have a `{ProjectPath}/ProjectSettings/ProjectVersion.txt` file which shows the version of Unity last used to open the project. We can leverage this to guess the installed Unity version in Unity Hub.

Since Unity Hub is a standard way to install Unity distributions, `.SetHubVersion()` is required for the majority of projects. With this auto-detection feature, it will be possible to omit this setup step.

- [x] Is based on my own work


matkoch:

PR #808, opened 2021-10-13
Target .net 6.0

ziongh:
To Target .Net Core 6.0 on any project, it needs only to add a global.json file to the root folder, and then the build.[sh,bat,ps1] will install the respective framework.

So actually this pull request is not at all a priority.

- [X] Is based on my own work


matkoch:
fixed on develop

PR #816, opened 2021-11-15
[Enhancement] Adds dotnet version, submodules, and fetch depth to GitHub Actions

ChaseFlorell:

- [x] Is based on my own work

In this pull request we are adding submodule support and dotnet version support to GitHub Actions.

Usage:

```csharp
[GitHubActions("foo",
    Submodules = GitHubSubmodules.Recursive,
    DotNetVersion = "6.0.x",
    IncludeDotNetPreRelease = true, // optional (defaults to false),
    FetchDepth = 2
 )]
```

Result:

```yaml
    steps:
      - uses: actions/checkout@v2
        with:
          submodules: recursive
          fetch-depth: 2
      - uses: actions/setup-dotnet@v1
        with:
          dotnet-version: 6.0.x
          include-prerelease: true
```

Supporting Documentation:  
 - https://github.com/actions/setup-dotnet  
 - https://github.com/actions/checkout


lahma:
We hit this issue while migrating NSwag to NUKE build: https://github.com/RicoSuter/NSwag/pull/3703 and my suggestion is to support multiple versions, I just monkey patched the support with custom attributes like you can see.

ChaseFlorell:
> We hit this issue while migrating NSwag to NUKE build: [RicoSuter/NSwag#3703](https://github.com/RicoSuter/NSwag/pull/3703) and my suggestion is to support multiple versions, I just monkey patched the support with custom attributes like you can see.

That is also what I did in my personal project in order to get around it. @matkoch is saying that we could accomplish the same with a `global.json` so that he doesn't have to support it in Nuke. I'm going to give that a try.

matkoch:
👋 @ChaseFlorell @lahma 

Submodules and fetch depth is very much okay for me to add to the attribute. With the .NET SDK version however, the reasoning is to push as much as possible to the common places, which in this case is – as @ChaseFlorell already said – the `global.json` file. Together with the bootstrapping files this should make sure that the required SDK is installed. If that's not the case, an issue would be nice. I can just say that my current build with NUKE on .NET 6 is working just as expected on all major CI services. So yeah, please let me know what's going on :)

matkoch:
Btw @lahma, cool to see that NUKE made it to NSwag and Quartz. I feel very sorry that I never managed to follow up with the PR that I promised 😓 But please always feel free to ping me wherever it's convenient for you 😊

lahma:
👋 @matkoch , thanks for chiming in!

> Submodules and fetch depth is very much okay for me to add to the attribute. With the .NET SDK version however, the reasoning is to push as much as possible to the common places, which in this case is – as @ChaseFlorell already said – the `global.json` file. Together with the bootstrapping files this should make sure that the required SDK is installed. If that's not the case, an issue would be nice. I can just say that my current build with NUKE on .NET 6 is working just as expected on all major CI services. So yeah, please let me know what's going on :)

Yes indeed the current bootstrapping does wonders and most definitely will work for 99% of the cases . What we hit is that when the single solution uses multiple targets, like old netcoreapp2.1, net3.1, net5.0, net6.0 for compatibility checking/regression testing (NSwag csprojs have a ton of conditional references) the build will fail:

https://github.com/RicoSuter/NSwag/runs/4166404439?check_suite_focus=true

```
It was not possible to find any compatible framework version
The framework 'Microsoft.NETCore.App', version '5.0.0' (x64) was not found.
  - The following frameworks were found:
      6.0.0 at [D:\a\NSwag\NSwag\.nuke\temp\dotnet-win\shared\Microsoft.NETCore.App]

You can resolve the problem by installing the specified framework and/or SDK.
```

From bootstrapping perspective it would indeed nice to just list all the required SDKs in `global.json`, so that they would be automatically available for people building on non-CI systems. I very much understand the reasoning here.

> Btw @lahma, cool to see that NUKE made it to NSwag and Quartz. I feel very sorry that I never managed to follow up with the PR that I promised 😓 But please always feel free to ping me wherever it's convenient for you 😊

Well don't feel sorry! I studied you example and adapted a slightly simpler approach as a start. I'm still very keen to explore the versioning and change log management you built, but for now I went with the minimal that works for Quartz. Your sample has helped me to become a NUKE advocate and I always try promote it when I can! (So far Spring.NET, Quartz.NET, NSwag on OSS side).

(The only thing so far that I have struggled has been negative globbing on top of default globbing, like everything from `src/**/*.cs` except for these three patterns. But as it's C# you can always get away with LINQ etc).

matkoch:
Yea, I saw the part about multiple SDKs already :) First of all, from my point of view, the global.json should be specifying the SDK used to make the build compile and run, and really _just_ the build. Whether other parts of the solution build or run correctly is a different matter. For that reason I'm leaning more towards installing the SDKs as part of the NUKE build. This also has the benefit that someone executing the full build locally will have the same experience (which they wouldn't taking the GitHub Actions workflow file approach). I'd also like to design this more straightforward, for instance by allowing an attribute on the build `InstallDotNetSdk(...)` or additional metadata in the project file. But for now I guess the easiest is to grab the `dotnet-install` file, and call it with all other required versions.

lahma:
Ah gotcha, sounds good 👍🏻

ChaseFlorell:
I'll try and find time to update this PR in order to remove the dotnet bits and the change in actions version.
I think there's probably value in allowing folks to opt-into the actions version in order to leverage the fetch depth.

matkoch:

PR #827, opened 2021-12-28
Add completer registrations to global tool

### LINKED ISSUE #597: Installation of shell completion
https://github.com/nuke-build/nuke/pull/827#issuecomment-1018172393

MartinSGill:

- [X] Is based on my own work

----

Resolve #597 

- New :RegisterCompleter command
- Simple implementation to echo code for supported shells
- Supports "invoke expression" style invocation,

e.g. for Powershell:

```powershell
(nuke ':RegisterCompleter' pwsh) -join "`n" | Invoke-Expression
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Added `PowerShellTasks`
 - Added `BootsTasks`
 - Added `NetlifyTasks`
+- Added `RegisterCompleter` command to global tool
 - Fixed check for executables compiled with `PublishSingleFile`
 - Fixed `MSBuild` localization using `MSBuildLocator`
 - Fixed Azure Pipelines caching
@@ -955,4 +956,3 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 [0.3.1]: https://github.com/nuke-build/nuke/compare/0.2.10...0.3.1
 [0.2.10]: https://github.com/nuke-build/nuke/compare/0.2.0...0.2.10
 [0.2.0]: https://github.com/nuke-build/nuke/tree/0.2.0
-
--- a/source/Nuke.Common/Utilities/CompletionUtility.cs
+++ b/source/Nuke.Common/Utilities/CompletionUtility.cs
@@ -13,6 +13,15 @@ namespace Nuke.Common.Utilities
 {
     public static class CompletionUtility
     {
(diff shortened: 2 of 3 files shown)


matkoch:
Hi @MartinSGill 

I'd probably like to aim for automatic installation, i.e., after confirmation, the global tool should write the completers to the appropriate files. Are you interested to investigate further?

MartinSGill:
@matkoch I can investigate further, although I'm not sure if automatic updating of files is a good idea, or easy to implement. I did consider it and felt it added too much complexity.

Choosing the file might be very tricky. 

For example in pwsh you'd have 2 choices: 

- `Microsoft.PowerShell_profile.ps1`
-  `profile.ps1` 

Or 4 choices if you assume people are using both Powershell 5 and 7. Do we put it in the PS5 profile or the PS7 profile or both?

For Bash I can think of three possible places to put the code: 
- `.profile`
- `.bashrc` 
- `.bash_profile`

I think the file used is mostly down to user preference, or whichever file the distro prefers.

This selection of files would also make it hard to detect if the auto-complete has already been added and to ensure not adding it a second time on subsequent runs.

Then there's the issue that many people heavily customise their profiles, even keeping them in git and using them across multiple machines. We'd never be able to detect and follow whatever style they prefer. So if we do go the automated route we'd still need to provide manual instructions for people (like me) that would not want a tool to modify their profile(s).

I feel providing instructions, such as "Add the following to your shell profile:"

```powershell
(nuke ':RegisterCompleter' pwsh) -join "`n" | Invoke-Expression
```

or

```bash
eval "$(nuke ':RegisterCompleter' bash)"
```

Might be least likely to upset users and also reduce future maintenance, and ease for adding new supported shells.

matkoch:
Thanks for the heads up.

I didn't put much thought into this until yet, but here is what I wish the user experience to look like:

1. User executes global tool for the first time. We detect the shell and suggest possible places for registrations.
2. User makes a choice, we add the snippets, and save a cookie under `%HOME%/.nuke`. This is created even if the user decides not to register.
3. Next run, the global tool would see the cookie and skip the registration setup. If no cookie exists, but it was added manually or through Git, we can still search all possible destinations if the snippets have already been registered – and create the cookie without asking.
4. The registration procedure could also decide whether to include the `nuke-` command, because we can check if session ids are available.

Last but not least, the greatest part about this is that we can also support completion for shared global tools. This has made me pretty excited once I realized it. 😃

matkoch:
Let's take a look later.

PR #850, opened 2022-01-24
Fix finding VS BuildTools 2022

ghost:
Fixes the search path for VS22 BuildTools, which seems to still install to Program Files (x86). 

The path on a docker container we have that uses vs_buildtools.exe to install VS BuildTools 2022:
image

Install Script:
```# MSbuildtools
RUN cd sources && \
    curl -SL --output vs_buildtools.exe https://aka.ms/vs/17/release/vs_buildtools.exe && \
    \
    (start /w vs_BuildTools.exe --quiet --wait --norestart --nocache \
    --add Microsoft.VisualStudio.Workload.MSBuildTools \
    --add Microsoft.VisualStudio.Workload.ManagedDesktopBuildTools \
    --includeRecommended --includeOptional \
    || IF "%ERRORLEVEL%"=="3010" EXIT 0) && \
    \
    del /q vs_buildtools.exe
```

- [X] Is based on my own work

### DIFF
--- a/source/Nuke.Common/Tools/MSBuild/MSBuildToolPathResolver.cs
+++ b/source/Nuke.Common/Tools/MSBuild/MSBuildToolPathResolver.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using JetBrains.Annotations;
 using Nuke.Common.Utilities;
 
 namespace Nuke.Common.Tools.MSBuild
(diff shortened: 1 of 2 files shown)


joergpichler:
This is also currently a blocker for us. Need this merged asap :)

dpisanu:
I approve of this fix. 👍

ghost:
Hey, @matkoch any way we could get this merged? I temporarily worked around this by installing full vs community to our build agents instead, but it does slow down our image builds quite a bit vs just installing the buildtools.

matkoch:
Hi, unfortunately, it's going to be a while. I'm currently occupied with other things.

MichaelKetting:
We also ran into this. There is the option of injecting the correct path so it's not a blocker, but still, is there something you need to help with releasing bugfixes in case an issue isn't as easily fixable?

PR #853, opened 2022-01-26
Added missing type name parameters for Kubernetes create and delete

SierraNL:
Added a type / name parameter to both Kubernetes create and delete parameters.
I think the current parameters where generated, and these ones are special, since they don't start with --.
So they are not generated, but currently you always need to specify them using `ConfigureProcessArguments`

As an example you currently can't create this
```bash
kubectl create namespace my-namespace
```
Without using `ConfigureProcessArguments`

This also goes for
```bash
kubectl delete namespace my-namespace
```

When it is added, it also enables you to combine things like:
```bash
kubectl delete pods --all
```
or
```bash
kubectl delete pods pod1,pod2
```

For more information:
https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#create
https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#delete

I created this PR as direct change to the json file. But I could not find the code that was originally used to generate this json with, neither could I find a clear name and description for the parameter besides this part of the documentation:
`kubectl delete ([-f FILENAME] | [-k DIRECTORY] | TYPE [(NAME | -l label | --all)])`
So if you want to have it implemented differently, please let me know.

- [X] Is based on my own work

### DIFF
--- a/source/Nuke.Common/Tools/Kubernetes/Kubernetes.json
+++ b/source/Nuke.Common/Tools/Kubernetes/Kubernetes.json
@@ -970,6 +970,13 @@
         "baseClass": "KubernetesToolSettings",
         "properties": [
           {
+            "name": "TypeName",
+            "type": "List<string>",
+            "format": "{value}",
+            "separator": " ",
+            "help": "The type (and name) of the resource. When only specifying a type either a --selector or --all needs to be specified"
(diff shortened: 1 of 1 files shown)


matkoch:

PR #887, opened 2022-03-12
Added attributes to supress build and target banners.

sharpninja:
1. `NoTargetBannerAttribute` - Attribute that is placed on a target declaration that suppresses the banner box only for that target.

2. `NoBuildBannerAttribute` - Attribute to be placed either on the Build class (which must be descended from `NukeBuild`) or placed on the assembly.  Its presence suppresses all stage output except what is written to Serilog.  It also does not affect the banners for Targets.

- [x] Is based on my own work

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -74,7 +74,7 @@ else {
     $env:DOTNET_EXE = "$DotNetDirectory\dotnet.exe"
 }
 
-Write-Output "Microsoft (R) .NET Core SDK version $(& $env:DOTNET_EXE --version)"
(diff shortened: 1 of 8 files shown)


bitbonk:
Nice idea. Shouldn't  this also be documented somewhere? How would a user discover the existence of these attributes?

sharpninja:
Happy to document it. As I've stated, I am totally open to changes. I haven't found a submission guide so would need to know what format to use and where to submit docs.

[Edit: removed clutter from email]

matkoch:
Sorry, but at the moment this is not something I'm interested in.

PR #894, opened 2022-03-20
Add support for Name and Parameters in Azure Pipelines

splaroche:
I had a personal need for slightly more expansive support for Azure Pipelines config generation, so I added in support for generating `Name` and `Parameters` to the `azure-pipelines.yml` file.  The parameters follows the pattern I found in the TeamCityAttribute.cs code.

Notes about parameters: 
- Only string, number, and boolean are supported.
  - As of right now, anything that doesn't fall into those will be set as 
string
- I had to introduce a new attribute to handle `values` in the `azure-pipelines.yml` spec.  These values turn the input into a radio button list in the UI, so it's a way to restrict values on a field to only what you want a CI user to be able to do in the CI.  
  - If applied to an enum type, and no value names are provided, the Enum values will be used instead.
- The parameters have to be passed into the command line call for nuke.  This is due to the way parameters are interpolated into the template (at run time), rather than being environment variables.

Let me know if there's any questions! 🙂 

- [x] Is based on my own work

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -74,7 +74,7 @@ else {
     $env:DOTNET_EXE = "$DotNetDirectory\dotnet.exe"
 }
 
-Write-Output "Microsoft (R) .NET Core SDK version $(& $env:DOTNET_EXE --version)"
+Write-Output "Microsoft (R) .NET SDK version $(& $env:DOTNET_EXE --version)"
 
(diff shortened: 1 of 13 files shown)


matkoch:
Doesn't align with the current plans, sorry.

PR #902, opened 2022-04-02
Add support for Bitbucket Pipelines host

lahma:
Created a Host implementation for [Bitbucket Pipelines](https://bitbucket.org/product/features/pipelines). I figured this is small and trivial enough to be suggested as PR instead of initial discussion. Ready for scrutiny 😉 

If this seems worthwhile addition I could also check what it would require to create YAML generation automation for basic usage.

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common/CI/Bitbucket/Bitbucket.cs
+++ b/source/Nuke.Common/CI/Bitbucket/Bitbucket.cs
@@ -1,0 +1,157 @@
+﻿// Copyright 2022 Maintainers of NUKE.
+// Distributed under the MIT License.
+// https://github.com/nuke-build/nuke/blob/master/LICENSE
+
+using System;
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using JetBrains.Annotations;
+using Nuke.Common.Utilities;
+
+namespace Nuke.Common.CI.Bitbucket
+{
+    /// <summary>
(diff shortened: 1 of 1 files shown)


matkoch:
Excellent pull-request :)

matkoch:
@lahma would just `Bitbucket` also work, or do you think `Pipelines` should also be in there? I notice that this might be a bit of an inconsistency. We have `AzurePipelines` and `GitHubActions`, but also `GitLab`. I guess I'm just unsure because `BitbucketPipelines` is very long as a name :)

lahma:
@matkoch sure thing! I force-pushed rename to just `Bitbucket`, both namespace and type name. I went through naming couple times in my head and decided to follow `GitHubActions` naming scheme, but the namespace `CI` and inheriting from `Host` and implementing `IBuildServer` does create some context already...

Maybe `GitHubActions` could also be shortened to service provider short name `GitHub` at some point.

lahma:
I also created [test repo on Bitbucket side](https://bitbucket.org/lahma/nuke/addon/pipelines/home#!/results/5) and added some output to see chosen host, host detection seems to work as expected (build failure seems to be unrelated to this change).

> NUKE Execution Engine version 1.0.0 (Linux,.NETCoreApp,Version=v6.0)
> NUKE Execution Host Nuke.Common.CI.Bitbucket.Bitbucket

PR #909, opened 2022-04-05
Added an extension point to customize logging.

jeffreycoxdoosan:
Added a LogExtensionAttribute class that can be used to override the console log template, or modify/replace the log configuration.

- [x ] Follows the contribution guidelines
- [x ] Is based on my own work
- [x ] Is in compliance with my employer


matkoch:
I'm sorry but this is not a direction I intent to go.

jeffreycoxdoosan:
Would you accept a pull request to just shut it off the double timestamps for Jenkins then? The reason I didn't propose that is because timestamps there  configurable on the Jenkins side, but maybe most people have it turned on?

Issue #913, opened 2022-04-13
DotNetTestSettingsExtensions.AddTeamCityLogger() causes unwanted delays


jwfx:
### Usage Information

NUKE 6.0.1 / SDK 6.0.202 / net6.0 / Win10 x64

### Description

A call to `DotNetTestSettingsExtensions.AddTeamCityLogger()` is taking a lot longer than expected in NUKE projects where multiple NuGet packages are referenced in .csproj.

### Reproduction Steps

1. Create a NUKE project that has multiple NuGet packages referenced it its .csproj
2. Add a build step that executes unit tests, like so
```csharp
    Target Test => _ => _
        .DependsOn(Build)
        .Executes(() =>
        {
            var tests = Solution.GetProject("MyApp.Tests");

            DotNetTasks.DotNetTest(settings => settings
                .SetProcessWorkingDirectory(RootDirectory / "test" / "MyApp.Tests")
                .SetProjectFile(tests?.Path)
                .SetConfiguration(Configuration)
                .When(TeamCity.Instance != null, settings => settings
                    .AddTeamCityLogger()
                    .AddLoggers("console;verbosity=normal")));
        });
```
4. Execute it in TeamCity environment

### Expected Behavior

The "Test" build step should immediately start after the previous step finishes.

### Actual Behavior

Depending of the number of referenced NuGet packages and the performance of the machine it could take a long time until the unit test execution actually starts.
In our concrete case the build server takes ~65 seconds to crawl through 120MB of NuGet packages before it continues.

### Regression?

I don't think so, from what I can tell the extension method looks like that since its inception.

### Known Workarounds

https://github.com/nuke-build/nuke/blob/develop/source/Nuke.Common/CI/TeamCity/DotNetTestSettingsExtensions.cs#L20

I believe `NuGetPackageResolver.GetLocalInstalledPackage()` should be called with `resolveDependencies` set to `false`, as in my understanding we're only interested in the path to the NuGet package itself here.

I've successfully tested this and the delay on the build server drops to ~12 seconds down from ~65.

matkoch:
Where is your `TeamCity.Dotnet.Integration` referenced?

jwfx:
In the `_build.csproj` like so

```
    <PackageReference Include="TeamCity.Dotnet.Integration" Version="1.0.28">
      <PrivateAssets>all</PrivateAssets>
      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
[...]

PR #825, opened 2021-12-22
Proof of concept: attribute to allow running a target in a docker container

matt-richardson:
This PR is a proof-of-concept, hacky hacky first draft demo of an attribute that allows a target to be executed inside a docker container. It is not production quality nor in a state where it should be merged.

The new attribute looks like
```
    [RunInDockerContainer("mcr.microsoft.com/dotnet/sdk:5.0", PlatformFamily.Windows)]
```

Annoyingly, this was all working yesterday, but today i'm getting 
```
07:51:35 [INF] Cleaning directory C:\Data\dev\Octopus\spikes\nuke\.nuke\temp\nukebuild...
07:51:35 [INF] > "C:\Program Files\dotnet\dotnet.exe" publish C:\Data\dev\Octopus\spikes\nuke\build\_build.csproj --configuration Release --output C:\Data\dev\Octopus\spikes\nuke\.nuke\temp\nukebuild --self-contained True --runtime win-x64 --verbosity Quiet --nologo
07:52:09 [ERR] Target RunTargetInWindowsDockerImageTest failed
System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. The system cannot find the file specified.
File name: 'Microsoft.Build, Version=15.1.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'
   at Nuke.Common.Execution.DockerExecutor.CopyPackageReferences()
   at Nuke.Common.Execution.DockerExecutor.Execute(NukeBuild build, ExecutableTarget target) in C:\Data\dev\Octopus\spikes\nuke\source\Nuke.Common\Execution\DockerExecutor.cs:line 46
   at Nuke.Common.Execution.BuildExecutor.Execute(NukeBuild build, ExecutableTarget target, IReadOnlyCollection`1 previouslyExecutedTargets, Boolean failureMode) in C:\Data\dev\Octopus\spikes\nuke\source\Nuke.Common\Execution\BuildExecutor.cs:line 112
```
I think it's environmental though, so still working through it.

_this pr desc is still a work-in-progress_

I confirm that the pull-request:

- [x] Follows the contribution guidelines
- [x] Is based on my own work
- [x] Is in compliance with my employer

PR #920, opened 2022-05-02
Improve GetGlobalInstalledPackage performance by reading NuGet metadata directly from nuspec file

jwfx:
Fixes DotNetTestSettingsExtensions.AddTeamCityLogger() causes unwanted delays #913

Currently `NuGetPackageResolver.GetGlobalInstalledPackage()` always reads NuGet metadata from the nupkg file itself. Depending of the amount of referenced NuGet packages and the amount of transitive dependencies this could cause delays as each package needs to be unzipped to access the contained nuspec file.

This change improves the performance for the `PackageReference` case, as the packages are unpacked on disk during restore and the nuspec file can be read directly.

NuGet packages referenced via `packages.config` can not benefit, as they are not unpacked on disk.

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common/Tooling/NuGetPackageResolver.cs
+++ b/source/Nuke.Common/Tooling/NuGetPackageResolver.cs
@@ -391,7 +391,11 @@ namespace Nuke.Common.Tooling
             public InstalledPackage(string fileName)
             {
                 FileName = fileName;
-                Metadata = new PackageArchiveReader(fileName).NuspecReader;
+
+                var directory = new DirectoryInfo(Path.GetDirectoryName(fileName));
+                Metadata = directory.GetFiles("*.nuspec").Length == 1
+                    ? new PackageFolderReader(directory).NuspecReader
+                    : new PackageArchiveReader(fileName).NuspecReader;
             }
 
             public string FileName { get; }


matt-richardson:
I also found slow behaviour in some work I'm doing. I wonder if the approach [here](https://github.com/nuke-build/nuke/pull/825/files#diff-dc2d64365e8e20c9f09950248cdf8f4a02a6a77cdb871e3edb0ee7e4d7cf8c06R70-R153) (which limits how many files are actually looked at) also improves your scenario, @jwfx?

jwfx:
Doesn't this approach break recursive searching?

matt-richardson:
> Doesn't this approach break recursive searching?

Can you expand? I cant see where the recursive search part that you're referring to?

jwfx:
Recursive might have been the wrong term, I meant nested or dependent packages.
If you filter out packages [here](https://github.com/nuke-build/nuke/blob/82deb98358a9776ebf3fc1f6863279790deb59a0/source/Nuke.Common/Tooling/NuGetPackageResolver.cs#L152), doesn't that prevent dependent packages from being checked [here](https://github.com/nuke-build/nuke/blob/82deb98358a9776ebf3fc1f6863279790deb59a0/source/Nuke.Common/Tooling/NuGetPackageResolver.cs#L165)?

matt-richardson:
Ahh, that's a good point. I will look closer 👍

matt-richardson:
> Ahh, that's a good point. I will look closer 👍

You are correct, @jwfx - my change would've broken it. Good catch! :)

I did find adding a simple cache to `GetGlobalInstalledPackage` was super helpful here - https://github.com/nuke-build/nuke/pull/825/files#diff-dc2d64365e8e20c9f09950248cdf8f4a02a6a77cdb871e3edb0ee7e4d7cf8c06R26.

PR #939, opened 2022-06-14
Add retry for dotnet restore in build.ps1

pellared:
## Why

Today we experienced a lot of issues when restoring NuGet package. Retrying minimizes the probability of transient errors connected with using remote resources.

## What

Add retry for `dotnet restore` in `build.ps1`.

## Checklist

- [x] Is based on my own work

## Remarks

> Follows the contribution guidelines

Where are they?

Should I mention the improvement in the `CHANGELOG.md`?

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -30,9 +30,18 @@ $env:NUKE_TELEMETRY_OPTOUT = 1
 # EXECUTION
 ###########################################################################
 
-function ExecSafe([scriptblock] $cmd) {
-    & $cmd
-    if ($LASTEXITCODE) { exit $LASTEXITCODE }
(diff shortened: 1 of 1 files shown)


matkoch:
We didn't experience these issues, so why change it? :) It would need to be applied to the template, and then also to the Bash script. But anyway, since this is sporadical, people can choose to do so themselves. Lastly, please star the project ;)

pellared:
> We didn't experience these issues, so why change it? :) It would need to be applied to the template, and then also to the Bash script. But anyway, since this is sporadical, people can choose to do so themselves. 

@matkoch  It may happen also when one has no stable Internet connection. Let me know if you want me to update the Bash script or prefer to abandon it (as a feature creep).

> Lastly, please star the project ;)

 Done 😄 ⭐

Issue #951, opened 2022-06-24
Allow overridable component settings via ApplyX


matkoch:
### Description

- Adapt `GetTarget` to support `InvocationExpression`
  ```
  (((configurator.Body as InvocationExpression).Expression as MemberExpression).Expression as ConstantExpression).Value
  ```
- Introduce new `ApplyX` method with `Expression<Configure<T>>` (no overload, since ambiguous)
- Add assertion in existing `Apply` method and check if `configurator.Target/Method` is declared by interface
- Use `GetAllMembers` to get quasi-override for original member similar as in `ValueInjectionAttributeBase.GetMemberValue<T>`
- Maybe base calls via `.Apply<T>(...)`

<img width="852" alt="JetBrains Rider-nuke-common – SettingsEntity Apply cs-2022-06-25-Y7ugiORf@2x" src="https://user-images.githubusercontent.com/5005566/175719120-5ed70efb-c17f-4d0d-afe9-bfdef7606db1.png">
<img width="1037" alt="JetBrains Rider-nuke-common – SettingsEntity Apply cs-2022-06-25-FJP1g6oM@2x" src="https://user-images.githubusercontent.com/5005566/175719136-8af3ea5c-fe57-4bb7-bc2b-97c23d9ed7ff.png">

### Usage Example

```code
interface ICompile : INukeBuild
{
    Target Compile => _ => _
        .Executes(() =>
        {
            var settings = new DotNetBuildSettings()
                .ApplyX(x => x.CompileSettings);
        });

    Configure<DotNetBuildSettings> CompileSettings => _ => _;
}

interface ICompileWithSomething : INukeBuild
{
    Configure<DotNetBuildSettings> ICompile.CompileSettings => _ => _
        .SetProperty("foo", "bar");
}

class Build : ICompileWithSomething
{
}
```

bitbonk:
With the help of @matkoch I managed to get this method working:

```csharp
public static TSettings Apply<TComponent, TSettings>(this TSettings settings, Expression<Func<TComponent, Configure<TSettings>>> configurator)
{
     var target = configurator.GetTarget();
     var memberInfo = configurator.GetMemberInfo();
     var config = memberInfo.GetValue<Configure<TSettings>>(target);
     return config.Invoke(settings);
}
```

Which would have been called like this:
```csharp
DotNetBuild(_ => _
    .Apply(CompileSettingsBase)
    .Apply((ICompileEx x) => CompileSettings));
```

[...]

PR #963, opened 2022-06-29
Introduce overridable Apply

bitbonk:
This is my attempt to understand and then solve #951 
This is just a temp PR for discussion.


bitbonk:
Oops, this PR was not  supposed to be public yet.

Issue #957, opened 2022-06-27
Attribute to disable default output


matkoch:
### Description

Add an attribute that disables logo, target headers, and summary without having to pass arguments (like `--no-logo`). Particularly for common tools not used in CI/CD this is good to have.

```
class DisableDefaultOutput : Attribute
{
    public DisableDefaultOutput(params DefaultOutputKind[] disabledOutputs)
    {
        // Assume empty => all
    }
}


public enum OutputKind
{
    Logo,
    Target,
    Summary
}
```

Adapt `NukeBuild.WriteLogo`, `NukeBuild.WriteTarget`, and `NukeBuild.WriteSummary` to check for attribute.

### Usage Example

```code
[Disable
class Build : NukeBuild
{
    Target Foo => _ => _
        .Executes(() => {});
}
```

PR #966, opened 2022-07-02
Introduce DisableDefaultOutputAttribute

bitbonk:
This will add #957

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common/Execution/Host.cs
+++ b/source/Nuke.Common/Execution/Host.cs
@@ -77,20 +77,9 @@ namespace Nuke.Common
             return false;
         }
 
-        internal virtual void WriteSummary(NukeBuild build)
+        protected internal virtual void WriteErrorsAndWarnings()
         {
-            WriteSevereLogEvents(Logging.InMemorySink.Instance.LogEvents);
-            WriteSummaryTable(build);
-
-            if (build.IsSuccessful)
-                WriteSuccessfulBuild(build);
-            else
-                WriteFailedBuild();
-        }
-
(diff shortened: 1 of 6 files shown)


source/Nuke.Common/Execution/Host.cs
```diff
@@ -47,21 +47,23 @@ protected internal void WriteLogo()
             Debug();
         }
 
-        protected internal virtual IDisposable WriteBlock(string text)
+        protected internal virtual IDisposable WriteBlock(string text, bool enableHeaderOutput = true)
```
bitbonk: I am not so sure if adding this parameter to each and every Host implementation is so nice.

source/Nuke.Common/NukeBuild.Output.cs
```diff
@@ -3,28 +3,44 @@
...
 {
+    using System.Linq;
+    using System.Reflection;
+    using Nuke.Common.Utilities;
+
     partial class NukeBuild
     {
+        protected bool IsOutputEnabled(DefaultOutputKind outputKind)
+        {
+            return GetType()
```
bitbonk, line 17: Should this be cached?

source/Nuke.Common/NukeBuild.Output.cs
```diff
@@ -3,28 +3,44 @@
...
+                .GetCustomAttributes<DisableDefaultOutputAttribute>()
+                .All(a => !a.IsOutputDisabled(outputKind));
```
bitbonk, line 19: Should this be cached?

PR #1031, opened 2022-10-11
Update versions of generated GitHub Actions tasks

esond:
This updates the versions of the generated cache and checkout tasks for GitHub Actions. The currently generated tasks are deprecated and result in a warning that reads something like:

> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/cache

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common/CI/GitHubActions/Configuration/GitHubActionsCheckoutStep.cs
+++ b/source/Nuke.Common/CI/GitHubActions/Configuration/GitHubActionsCheckoutStep.cs
@@ -1,4 +1,4 @@
-// Copyright 2021 Maintainers of NUKE.
+﻿// Copyright 2021 Maintainers of NUKE.
 // Distributed under the MIT License.
 // https://github.com/nuke-build/nuke/blob/master/LICENSE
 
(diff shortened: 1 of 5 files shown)


Simonl9l:
FYI there is current a bug in [checkout](https://github.com/actions/checkout/issues/882) that means its not correctly pulling tags when FetchDepth:0.

There is a [workaround](https://github.com/actions/checkout/issues/882#issuecomment-1231911465) not sure how this could be factored in.

Simonl9l:
Just popped up on GitHub Workflows... 
image

source/Nuke.Common/CI/GitHubActions/Configuration/GitHubActionsCheckoutStep.cs
```diff
@@ -17,7 +17,7 @@ public class GitHubActionsCheckoutStep : GitHubActionsStep
 
         public override void Write(CustomFileWriter writer)
         {
-            writer.WriteLine("- uses: actions/checkout@v2");
```
lahma, line 20: Might be beneficial for the `actions/checkout@v2` part to be public get/set property to allow people to override in their own pipelines when needed if yet another update comes in? 

It would then be just a matter of finding with type and then setting the value to something else.

lahma: Same goes for the `actions/upload-artifact@v3` too if that makes sense.

PR #1065, opened 2022-11-25
Update Azure Pipelines Publish step

lg2de:
I'm currently working on implementing support for Azure DevOps **Server**.
While I was doing that, I found small issue which also apply to DevOps **Services**:
1. `pathtoPublish` is a small typo, `pathToPublish` is correct
2. If you have multiple targets creating artifacts, the generated `PublishBuildArtifacts@1` always have the same display name. So, I've introduced new `DisplayName` property which is build from artifacts name.

- [X] Is based on my own work


lg2de:
I would like understand why do you removed the display name, @matkoch .

matkoch:
Please only 1 change per pull-request. I applied similar changes in another commit https://github.com/nuke-build/nuke/commit/394cee98b05a1e6a08ea7a6ea9b796b987ebe96e

lg2de:
This is why I created two separate commits. But ok, next time I will create two PRs.
Thanks for your great project!

PR #1068, opened 2022-12-02
Expose ExitCode on ProcessException

### LINKED ISSUE #875: Make available exit code of failed process (ProcessException.Process)
### Description

When I use custom tool lightway integration, the only way how to handle errors in process is catching `ProcessException`. But just to catch the exception is often not enough. I would like to have some logic and decide, if there is really a problem, or not. The minimum is to have the value of process exit code and decide what to do. Because some exit codes are not real errors and can be handled and the build can continue.

### Usage Example

``` cs
[PathExecutable("dotnet-affected")]
readonly Tool DotnetAffected;

// In target execution:
try
{
    DotnetAffected(args);
}
catch (ProcessException ex)
{
    // We use this kind of logic.
    // 'GetProcessExitCode()' is our extension method which uses reflection to pull out exit code.
    if (ex.GetProcessExitCode() == SomeOkExitCode)
    {
        // I can handle this.
    }
    else
    {
        // Anything else is something I cannot handle.
        throw;
    }
}
```

### Alternative

I use reflection to pull out exit code of the process from `ProcessException`.

Baune8D:
* Expose the exit code on `ProcessException`.

It would be immensly useful to be able to get the actual exit code instead of parsing the error message.

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common/Tooling/ProcessException.cs
+++ b/source/Nuke.Common/Tooling/ProcessException.cs
@@ -54,5 +54,7 @@ namespace Nuke.Common.Tooling
         }
 
         internal IProcess Process { get; }
+
+        public int ExitCode => Process.ExitCode;
     }
 }


ron-myers:
Good work @Baune8D - I have wanted this a few times in the past.

matkoch:
Next version is going to have this

```
DotNetRestore(_ => _
    .SetProjectFile(Solution)
    .DisableProcessAssertZeroExitCode());

DotNetRestore(_ => _
    .SetProjectFile(Solution)
    .SetProcessCustomExitHandler((_, p) =>
        p.ExitCode switch
        {
            0 => null,
            _ => throw new Exception()
        }));

DotNetRestore(_ => _
    .SetProjectFile(Solution + "aaaa")
    .SetProcessCustomExitHandler(p =>
    {
        switch (p.ExitCode)
        {
            case 1:
                break;
        }
    }));
```

PR #1075, opened 2022-12-12
Support packed-refs in GitRepository

### LINKED ISSUE #795: Support packed-refs in GitRepository
### Description

In Git, refs can be packed with [git pack-refs](https://git-scm.com/docs/git-pack-refs) which creates a single `.git/packed-refs` file with all the refs. According to [documentation](https://git-scm.com/docs/git-pack-refs):

> This command is used to solve the storage and performance problem by storing the refs in a single file, $GIT_DIR/packed-refs. When a ref is missing from the traditional $GIT_DIR/refs directory hierarchy, it is looked up in this file and used if found.

Currently, Nuke only looks for refs in `refs` directory
https://github.com/nuke-build/nuke/blob/217b8080c3011c4b06dbe6953b837ce5ce663d7a/source/Nuke.Common/Git/GitRepository.cs#L98-L106

Would be cool to have it also look into `packed-refs`

### Usage Example

_No response_

### Alternative

_No response_

lahma:
Adds support for determining the commit information from `packed-refs` file. Tested by calling `git gc` in nuke repo which broke the commit finding logic, after implementing fix, worked.

fixes #795

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Build/VCS/GitRepository.cs
+++ b/source/Nuke.Build/VCS/GitRepository.cs
@@ -99,9 +99,19 @@ namespace Nuke.Common.Git
             if (!head.StartsWith("refs/heads/"))
                 return head;
 
-            var headRefFile = gitDirectory / head;
-            Assert.FileExists(headRefFile);
-            return headRefFile.ReadAllLines().First();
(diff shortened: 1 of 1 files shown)


matkoch:

PR #1086, opened 2022-12-20
Fix `ExitCode` cannot be accessed because the underlying process has been detached

Baune8D:
When catching `ProcessException`, `ExitCode` cannot be accessed because the underlying process seems to be detached. Fix this by saving `ExitCode` on `ProcessException` creation.

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Tooling/ProcessException.cs
+++ b/source/Nuke.Tooling/ProcessException.cs
@@ -45,6 +45,7 @@ namespace Nuke.Common.Tooling
             : base(FormatMessage(process))
         {
             Process = process;
+            ExitCode = process.ExitCode;
         }
 
         protected ProcessException(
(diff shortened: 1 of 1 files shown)


matkoch:
I don't really understand why this is necessary. What is the exception message? And wouldn't this apply to all properties that can be accessed on the process?

Baune8D:
> I don't really understand why this is necessary. What is the exception message? And wouldn't this apply to all properties that can be accessed on the process?

I seem to be getting the following exception when trying to access the `ExitCode` property:
```
System.InvalidOperationException: No process is associated with this object.
   at System.Diagnostics.Process.EnsureState(State state)
   at System.Diagnostics.Process.get_HasExited()
   at System.Diagnostics.Process.EnsureState(State state)
   at Sys…
```
This is tested by catching an `AggregateException` and accessing `ExitCode` through `InnerException` 

image

PR #1120, opened 2023-01-25
Fix ExtractChangelogSectionNotes not throwing the intended exception

sandrofigo:
The `ExtractChangelogSectionNotes()` method if provided with a tag string that does not exist in the changelog throws an exception with the message `"Sequence contains no matching element"`, instead of the intended exception message `"Could not find release section for '{tag}'."` defined here:
https://github.com/nuke-build/nuke/blob/6506f3081926e89232e1811b77f5c984a7bf17b5/source/Nuke.Common/ChangeLog/ChangeLogTasks.cs#L172

Consider the changelog to look similar to this:
```markdown
# Changelog
## [Unreleased]
### Added
- Some text that spans
  multiple lines
- A text on a single line

## [0.2.3] - 2023-01-02
### Fixed
- Fixed something
```
Calling the method with a non-existing tag (in this case "4.0.4")
```csharp
ChangelogTasks.ExtractChangelogSectionNotes(<pathToChangelogFile>, "4.0.4")
```
would result in an exception with the message `"Sequence contains no matching element"` instead of `"Could not find release section for '4.0.4'."`

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common.Tests/SettingsTest.cs
+++ b/source/Nuke.Common.Tests/SettingsTest.cs
@@ -30,18 +30,22 @@ namespace Nuke.Common.Tests
         [Fact]
         public void TestCommon()
         {
+            var logEntry = default((OutputType Type, string String));
             var settings = new DotNetRunSettings()
                 .SetProcessToolPath("/path/to/dotnet")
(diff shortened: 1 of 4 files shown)


sandrofigo:
Moved to #1130

PR #1131, opened 2023-02-13
Change type of `JobId` from `int` to `string` at coveralls.net tool

### LINKED ISSUE #1116: Use `long?` (or `string`)  for `JobId` in `Tools.CoverallsNet`
### Description

When passing the `GitHubActions.RunId` as `JobId`.. that does not work anymore.. since the run id is beyond `int.MaxValue`.

image

### Usage Example

_No response_

### Alternative

_No response_

Maybe.. How does the `CoverallsNet.Generated.cs` generates? Do I only need to change `int` to `long` and rebuild?

ITaluone:
Fixes #1116 

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common/Tools/CoverallsNet/CoverallsNet.json
+++ b/source/Nuke.Common/Tools/CoverallsNet/CoverallsNet.json
@@ -112,7 +112,7 @@
           },
           {
             "name": "JobId",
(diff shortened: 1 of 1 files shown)


ITaluone:
Sorry.. closed accidentally :)

PR #1137, opened 2023-03-09
Update ITargetDefinition.cs

VladyslavLishchyna:
add property Name to ITargetDefinition

class TargetDefinition already have "Name" property so I decided also add it to ITargetDefinition because library users can work only with ITargetDefinition and name very helpful to expend custom functionality. 


matkoch:
Not fond of this because it leaks into code completion on user level

PR #1141, opened 2023-03-14
Ensured that the SonarQube verbose flag is lower-cased

### LINKED ISSUE #874: Sonar EnableVerbose dosen't work
### Usage Information

6.0.1 / .NET 6

### Description

When enabling sonar verbose get the following warning and sonar doesn't output verbose
`[DBG] WARNING: Expecting the sonar.verbose property to be set to either 'true' or 'false' (case-sensitive) but it was set to 'True'.`

### Reproduction Steps

```
sonarScannerBeginSettings.EnableVerbose()
```

### Expected Behavior

Sonar to output verbose logs

### Actual Behavior

Verbose logs are not shown

### Regression?

_No response_

### Known Workarounds

_No response_

dennisdoomen:
Fixes #874

Also includes a proposal on how to test such a thing. Please let me know what you think of that.

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common.Tests/Tools/SonarScannerTasksTests.cs
+++ b/source/Nuke.Common.Tests/Tools/SonarScannerTasksTests.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Linq;
+using Nuke.Common.Tools.SonarScanner;
+using Xunit;
+using FluentAssertions;
+
+namespace Nuke.Common.Tests.Tools
+{
+    public class SonarScannerBeginSettingsTests
+    {
+        [Fact]
+        public void Verbose_logging_needs_lowercase_flag()
+        {
+            // Act
+            var settings = new SonarScannerBeginSettings().EnableVerbose();
+
+            // Assert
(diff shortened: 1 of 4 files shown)


matkoch:
Isn't this (lower-case boolean) also relevant for all other properties that are boolean?

dennisdoomen:
> Isn't this (lower-case boolean) also relevant for all other properties that are boolean?

Could be. But this one triggered a warning/error in the SQ logs.

matkoch:
Could you check with another parameter perhaps? If it's not working, then I'd like to fix the other parameters as well. If it's working, then it's rather a bug for SonarQube.

dennisdoomen:
I don't have access to a SQ environment for the next couple of weeks (because I switched clients), but it's very likely that it's going to able to other boolean parameters as well.

matkoch:
There's another flag to write less code. Committing manually, but thanks for bringing this up.

dennisdoomen:
> There's another flag to write less code. Committing manually, but thanks for bringing this up.

What do you mean with that?

matkoch:
`customValue` instead of `customImpl`

PR #1160, opened 2023-05-06
Add GitHubActionsAttribute.CheckoutTokenSecret

olegtarasov:
I've added an option to specify GitHub token at the checkout stage when generating GitHub Actions yaml file. This enables checking out submodules which are private GitHub repositories. In order to discourage setting this token in plain text, it's only possible to set it through a secret.

More info here: https://github.com/orgs/community/discussions/25516#discussioncomment-3248190

- [x] Is based on my own work


matkoch:
I don't think we need a new env var for this. GitHub already populates `GITHUB_TOKEN` for each workflow, so I think that could be overwritten. If that's the case, we can add the checkout token unconditionally whenever `submodules` is true.

PR #1162, opened 2023-05-08
Fix console detection on Windows 10+

### LINKED ISSUE #1161: Banner, target names etc. seem to have the wrong color when using a dark console theme
### Usage Information

7.0.0

### Description

Banner, target names etc. seem to have the wrong color when using a dark console theme.

In my case they are have the same color as the background and are thus invisible:
image

When switching to a light theme they get visible:
image

### Reproduction Steps

In my case, just run a build.

### Expected Behavior

Colors should be visible.

### Actual Behavior

See description.

### Regression?

This worked perfectly fine in 6.3.

Screenshot from a different project using v 6.3.
image

### Known Workarounds

_No response_

No

lahma:
Detect Windows 10+ as ANSI output capable and fix `SystemConsoleHostTheme` `Debug` level color

fixes #1161

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Build/Logging.cs
+++ b/source/Nuke.Build/Logging.cs
@@ -6,6 +6,7 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Runtime.InteropServices;
 using JetBrains.Annotations;
 using Nuke.Common.CI;
 using Nuke.Common.Execution.Theming;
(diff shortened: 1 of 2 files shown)


batzen:
The correct way to check if ansi support is available in Windows should be something like:

https://github.com/spectreconsole/spectre.console/blob/3ec0fee795b641db6fc1780f09c34312e4e73f56/src/Spectre.Console/Internal/Backends/Ansi/AnsiDetector.cs#L91

The problem on Windows, even 10+, is that the default console host does not support ansi/VT.
That support is delivered by console hosts like WT (Windows Terminal).

lahma:
Might be just easier to remove my try to detect ANSI console support and go with the basic terminal version on Windows.

matkoch:
Reinstantiated the previous behavior. Thanks anyway for the PR!

PR #1178, opened 2023-05-18
Use string.Replace to redact secrets

### LINKED ISSUE #1169: Can't filter invocation
### Usage Information

NUKE 7.0.0

### Description

In NUKE 6.3.0 I used to be able to use the `outputFilter` to redact **both** the invocation and the output for a `DotNet` command. In NUKE 7.0.0, this parameter is gone and `logger` is only capturing the output.

### Reproduction Steps

In NUKE 6.3.0:

```csharp
DotNet(
    $"user-secrets set {secretName} {secretValue} --id {UserSecretsId}",
    outputFilter: o => o.Replace(secretValue, "*****", StringComparison.Ordinal));
```

In NUKE 7.0.0:

```csharp
DotNet($"user-secrets set {secretName} {secretValue} --id {UserSecretsId}", logger: RedactLog, logInvocation: false);

void RedactLog(OutputType type, string s)
{
    var redactedLog = s.Replace(secretValue, "*****", StringComparison.Ordinal);
    Serilog.Log.Write(type == OutputType.Err ? LogEventLevel.Error : LogEventLevel.Information, redactedLog);
}
```

### Expected Behavior

These are the logs I used to get in NUKE 6.3.0:

```plaintext
07:10:57 [INF] > /usr/share/dotnet/dotnet user-secrets set ServiceBusConnection ***** --id 074ca336-270b-4832-9a1a-60baf152b727
07:10:57 [DBG] Successfully saved ServiceBusConnection = ***** to the secret store.
```

### Actual Behavior

These are the logs I'm getting in NUKE 7.0.0 with my workaround:

```plaintext
22:05:16 [INF] Successfully saved ServiceBusConnection = ***** to the secret store.
```

### Regression?

The parameter `outputFilter` used to be present in NUKE 6.3.0. I don't know if this is regression as the name of the parameter suggests it should only have been filtering the output. But I came to rely on this behaviour.

### Known Workarounds

As a partial workaround, I can use the `logger` parameter but it doesn't give me the opportunity to redact the invocation.

No

gabrielweyer:
I'm now using `string.Replace` to redact secrets instead of `ReplaceRegex`. Secrets can potentially contain regular expression metacharacters which leads to unpredictable behaviour.

Fixes #1169.

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Tooling/ProcessTasks.cs
+++ b/source/Nuke.Tooling/ProcessTasks.cs
@@ -79,7 +79,7 @@ public static class ProcessTasks
             static Func<string, string> GetOutputFilterForArgumentStringHandler(ref ArgumentStringHandler arguments)
             {
                 var redactedValues = arguments.SecretValues;
-                return x => redactedValues.Aggregate(x, (arguments, value) => arguments.ReplaceRegex(value, _ => Arguments.Redacted));
+                return x => redactedValues.Aggregate(x, (arguments, value) => arguments.Replace(value, Arguments.Redacted));
             }
 
             return StartProcess(


matkoch:
Fixed for 7.0.2. It shouldn't use regex replace.

matkoch:
Didn't see the update, but it's the same change indeed.

PR #1021, opened 2022-09-28
WIP on-prem agent Azure pipeline pool logic

MarkZither:
This PR addresses #1019, I added a new AzurePipelineImage of OnPrem and a string array property `OnPremPools`.

The code in AzurePipelinesAttribute is a bit of copy and paste of GetStage, GetJob and GetSteps to allow the pool name to be passed as a string rather than passing the AzurePipelineImage and the nested ternary in AzurePipelinesCacheStep AdjustedPath needs improving, but it is currently working as I hoped.

Input like
``` 
[AzurePipelines(
    suffix: null,
    AzurePipelinesImage.OnPrem,
    OnPremPools = new[] { "my-onprem-pool" },
```
give me the output I want
```
stages:
  - stage: my_onprem_pool
    displayName: 'my-onprem-pool'
    dependsOn: [  ]
    pool: 'my-onprem-pool'
```

I confirm that the pull-request:

- [ x ] Follows the contribution guidelines
- [ x ] Is based on my own work
- [ x ] Is in compliance with my employer

PR #1185, opened 2023-05-30
Adding self-hosted option to GitHubActionsImage

MarioGK-AutoSystem:
Just added to SelfHosted to GitHubActionsImage enum

- [ ] Follows the contribution guidelines
- [ ] Is based on my own work
- [ ] Is in compliance with my employer


lahma:
This is a nice addition. We opted to creating inherited attribute that allows setting `runs-on` more precisely,  like `runs-on: [self-hosted,linux]` (needs to be on premises AND linux machine). While image enumeration is nice it could also beneficial to have some sort of virtual method maybe that returns the string presentation or something similar to allows more fine-grained configuration.

matkoch:
I'm not using self-hosted agents, but I feel the same way as commented here:

https://github.com/nuke-build/nuke/pull/1021#issuecomment-1315657693

psimsa:
`self-hosted` is a valid built-in Github Actions option though, unlike OnPrem in #1021. There are many good reasons to use self-hosted runners, especially in private repos...

PR #1188, opened 2023-05-31
Fix #1186 Fixing signtool.exe path for new windows kits

### LINKED ISSUE #1186: "SignToolTasks.SignToolPath" can't find the correct path for signtool.exe
### Usage Information

Nuke.Common 7.0.2 / Microsoft.NET 4.8.1 / Windows 10

### Description

When trying to run "SignToolTasks.SignTool()", I have the following error:

> 
> System.ArgumentException: Expected condition to be true (Parameter 'toolPath != null')
> at Nuke.Common.Assert.True(Boolean condition, String message, String argumentExpression) in /_/source/Nuke.Common/Assert.cs:line 34

"SignToolTasks.SignToolPath" looks for an old array of paths that don't correspond anymore with the latest windows kits released.
(current code: https://github.com/nuke-build/nuke/blob/develop/source/Nuke.Common/Tools/SignTool/SignToolSettings.cs)

For instance, I have "signtool.exe" installed on the path: "C:\Program Files (x86)\Windows Kits\10\bin\10.0.22621.0\x64\signtool.exe" (and the env variable "SIGNTOOL_EXE" not set)

### Reproduction Steps

Install any Windows Kit 10.*

### Expected Behavior

signtool should work without problems

### Actual Behavior

An exception is thrown

### Regression?

_No response_

### Known Workarounds

Set manually the env variable "SIGNTOOL_EXE" with the correct path.

Yes

gustavocalheiros:
I look inside the path of the latest Windows Kit major version (in this case "10") and check in all the minor version folders (for example "10.0.22621.0", "10.0.16299.0"), and get the last one that contains the file "signtool.exe".

if nothing is found, I fallback on the previous code.

- [x] Is based on my own work


gustavocalheiros:
Fix #1186

matkoch:
Could you unify the code to only make use of `GlobFiles` ?

gustavocalheiros:
yep, will do it in the next days.

gustavocalheiros:
@matkoch  do you mind if we drop the support to old versions of the Windows Kit (I mean support only 10.*) ?

PS: Windows Kit 10 is the current version of the SDK for windows 10 and 11.

matkoch:
fine i guess

gustavocalheiros:
done

gustavocalheiros:
@matkoch done

gustavocalheiros:
hello @matkoch, any chance this PR will be merged soon?

source/Nuke.Common/Tools/SignTool/SignToolSettings.cs
```diff
@@ -20,18 +19,12 @@ private static string GetToolPath()
...
-                   programDirectory / "Windows Kits" / "8.1" / "bin" / platformIdentifier,
-                   programDirectory / "Windows Kits" / "8.0" / "bin" / platformIdentifier,
-                   programDirectory / "Microsoft SDKs" / "Windows" / "v7.1A" / "Bin"
-               }
-            .Select(x => x / "signtool.exe")
-            .WhereFileExists()
-            .FirstOrDefault();
+        var windowsKitsRootDirectory = programDirectory / "Windows Kits" / windowsKitLastVersion;
+
+        var signToolPath = windowsKitsRootDirectory.GlobFiles($"bin/{windowsKitLastVersion}.*/{platformIdentifier}/signtool.exe").LastOrDefault();
```
MichaeIDietrich: Shouldn't we order by version number first before using `LastOrDefault`?

MichaeIDietrich:
What do you think about adding support for using the `signtool.exe` from these two NuGets?
https://www.nuget.org/packages/Microsoft.Windows.SDK.BuildTools
https://www.nuget.org/packages/signtool

source/Nuke.Common/Tools/SignTool/SignToolSettings.cs
gustavocalheiros: the order returned is naturally the alphabetical order, meaning version 10.0.22 will be after version 10.0.17.

gustavocalheiros:
> What do you think about adding support for using the `signtool.exe` from these two NuGets? https://www.nuget.org/packages/Microsoft.Windows.SDK.BuildTools https://www.nuget.org/packages/signtool

@MichaeIDietrich  Do you think it is possible to the user to run nuke-build having no SDK kit installed?
if yes, I'm ok to add support only to the 1st one (official microsoft package). What do you think?

gpailler:
Hello @gustavocalheiros, I would say yes. The Microsoft SDK is not a prerequisite for Nuke Build

source/Nuke.Common/Tools/SignTool/SignToolSettings.cs
MichaeIDietrich: Thanks for the info, I did not know it's already ordered. 👍

MichaeIDietrich:
Yes, the Windows SDK is **not** needed for Nuke to run. Only a matching .NET SDK is needed to run Nuke.

gustavocalheiros:
@MichaeIDietrich  Pushed support for [Microsoft.Windows.SDK.BuildTools](https://www.nuget.org/packages/Microsoft.Windows.SDK.BuildTools)

source/Nuke.Common/Tools/SignTool/SignToolSettings.cs
```diff
@@ -20,11 +21,25 @@ private static string GetToolPath()
...
-        var signToolPath = windowsKitsRootDirectory.GlobFiles($"bin/{windowsKitLastVersion}.*/{platformIdentifier}/signtool.exe").LastOrDefault();
+        var signToolPath = windowsKitsRootDirectory.GlobFiles($"bin/{windowsKitVersionWildcard}/{platformIdentifier}/{signtoolExe}").LastOrDefault();
 
-        return signToolPath ?? windowsKitsRootDirectory.GlobFiles("App Certification Kit/signtool.exe").SingleOrDefault();
+        if(signToolPath == null)
+        {
+            var nugetPath = SettingsUtility.GetGlobalPackagesFolder(Settings.LoadDefaultSettings(null));
+
+            signToolPath = AbsolutePath.Create(nugetPath)
+                .GlobFiles($"{microsoftBuildToolsNugetPackage}/{windowsKitVersionWildcard}/bin/{windowsKitVersionWildcard}/{platformIdentifier}/{signtoolExe}")
```
gustavocalheiros, line 37: path structure for nuget packages is different, for example :
C:\Users\MyUser\ .nuget\packages\microsoft.windows.sdk.buildtools\10.0.20348\bin\10.0.20348\x64\signtool.exe

gpailler:
@MichaeIDietrich @matkoch 
Do you think we should align with the approach taken by other packages, such as XUnit and Codecov, by updating `SignTool.json` with:
```
  "nugetPackageId": "Microsoft.Windows.SDK.BuildTools",
  "customExecutable": true,
```
and implementing the `GetProcessToolPath()` method? 
This would be a breaking change, but it seems to be more in line with the conventions used by other tools.

ITaluone:
@gustavocalheiros Do you mind fixing this in: https://github.com/Fallout-build/Fallout ?

PR #1182, opened 2023-05-21
Fix #1181 UnityTasks.Unity throws exception on macOS

chrisyarbrough:
This fixes #1181.

When executing the Unity tool on macOS, an exception was thrown because the FilePathExtension was not able to determine if `hubToolPath` was a file or directory. The check is now explicit and does not rely on the naming of the variable.

I considered adding tests for this issue, but found that it would require sweeping changes to decouple the platform and system logic. A more simplified test would be one that would only detect the issue when actually running on macOS and not on Windows, but I don't yet know if the Nuke project has anything set up for platform testing to make such tests feasible.

For now, the change is simple enough to warrant that we go without a test, I believe.

I confirm that the pull-request:

- [x] Follows the contribution guidelines
- [x] Is based on my own work
- [x] Is in compliance with my employer

Issue #1181, opened 2023-05-21
UnityTasks.Unity throws exception in AbsolutePathExtensions.Exists on macOS


chrisyarbrough:
### Usage Information

Nuke.Common 7.0.2 / Microsoft.NET.Sdk 6.0 / macOS

### Description

Running UnityTasks.Unity on macOS worked correctly in Nuke version 6.3.0, but throws an exception starting in 7.0.0 (and still in 7.0.2). Sample code:

```
Target Test => _ => _
	.Executes(() =>
	{
		// This works in Nuke.Common 6.3.0, but throws an exception in 7.0.2 on macOS:
		// System.ArgumentException: Cannot infer from argument 'hubToolPath' if either file or directory must exist
		// at Nuke.Common.IO.AbsolutePathExtensions.Exists(AbsolutePath path, String expression) in /_/source/Nuke.Utilities/IO/AbsolutePath.Exists.cs:line 29
		UnityTasks.Unity(settings => settings
			.EnableBatchMode()
			.SetProjectPath(RootDirectory)
			.EnableRunEditorTests()
			.SetLogFile(ArtifactsDirectory / "Unity.log")
			.SetEditorTestsResultFile(ArtifactsDirectory / "TestResults.xml")
		);
	});
```

### Reproduction Steps

Open the provided repro project:

[NukeBugReport_UnityTasks_AbsolutePathExtensions_Exception.zip](https://github.com/nuke-build/nuke/files/11524778/NukeBugReport_UnityTasks_AbsolutePathExtensions_Exception.zip)

And run `nuke test`.

### Expected Behavior

The simple UnityTask should run without errors.

### Actual Behavior

An exception is thrown:

```
[ERR] Target Test has thrown an exception
System.ArgumentException: Cannot infer from argument 'hubToolPath' if either file or directory must exist
   at Nuke.Common.IO.AbsolutePathExtensions.Exists(AbsolutePath path, String expression) in /_/source/Nuke.Utilities/IO/AbsolutePath.Exists.cs:line 29
   at Nuke.Common.Tools.Unity.UnityTasks.DetectUnityVersion(UnitySettings& unitySettings) in /_/source/Nuke.Common/Tools/Unity/UnityTasks.cs:line 81
   at Nuke.Common.Tools.Unity.UnityTasks.PreProcess(UnitySettings& unitySettings) in /_/source/Nuke.Common/Tools/Unity/UnityTasks.cs:line 70
[...]

PR #1193, opened 2023-06-02
Fix Unity detection code

kkopanski:
I have fixed `DetectUnityVersion` which was failing due to incorrectly named variable.
`AbsolutePath.Exists` requires some special naming to work.

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common/Tools/Unity/UnityTasks.cs
+++ b/source/Nuke.Common/Tools/Unity/UnityTasks.cs
@@ -78,15 +78,15 @@ public partial class UnityTasks
             return;
 
         var editorVersion = ReadUnityEditorVersion(unitySettings.ProjectPath);
-        var hubToolPath = GetToolPathViaHubVersion(editorVersion);
-        if (hubToolPath.Exists())
+        var hubToolExe = GetToolPathViaHubVersion(editorVersion);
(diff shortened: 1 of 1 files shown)


kkopanski:
Just found an related issue (#1181) with different resolution.

chrisyarbrough:
Yea, I mentioned it on my PR, but I would recommend being more explicit about how these things work. Rather than relying on the naming of a variable, better use a dedicated method to check for a file versus a folder. I feel like this naming-convention thing will break again soon in the future if people refactor. It also limits how you can pick good names, and that's hard enough on its own. For example, it makes no sense to call the variable "exe" because the Unity Hub is not and "exe" on macOS, it's an "app". But names like "UnityHubFilePath" and "UnityInstallationDirectoryPath" are pretty accurate.

But hey, thanks for helping with Nuke! Let's see what the maintainer or others think about the topic. :)

kkopanski:
Hey @chrisyarbrough!
First of all thank for your immediate response. 

To be honest I’m totally commited with your solution. 
I’m pretty new in NUKE and just missed that this metod exists. 
Wherever we know for what we are looking for we should use explicit method for it and not depend on variable name. 

This PR should be closed in favor of #1182

PR #1194, opened 2023-06-02
Add Unity Test Framework support

kkopanski:
Rework of not working test running command line options.
AFAIK this never works on any Unity version.

Also added new `TestPlatform` support.

- [x] Is based on my own work


chrisyarbrough:
Just to add some info:
In Unity 5.3, there were different CLI arguments, e.g. "runEditorTests". See: https://docs.unity3d.com/530/Documentation/Manual/CommandLineArguments.html

But @kkopanski has made the changes to work with the more recent Test Framework:
https://docs.unity3d.com/Packages/com.unity.test-framework@1.1/manual/reference-command-line.html

I personally don't care too much for legacy versions and it's probably best to align with the official support cycle of Unity.
https://unity.com/releases/editor/qa/lts-releases

So in this case, since the oldest LTS version is now 2021 it's fine.
The Test Framework added the new commandline arguments with 1.0: https://docs.unity3d.com/Packages/com.unity.test-framework@1.0/manual/reference-command-line.html

Which requires Unity 2019.2 or later, so it's been changed a long time now.

kkopanski:
@chrisyarbrough you are totally right.
I have totally missed the 5.3 and lower versions of Unity.

I will rework this code so older methods will just stay as is and add new ones.

We should also talk about marking things as deprecated if they are below Unity support cycle.
What do you think?

kkopanski:
Just added new UnityProjectSettings class with all project-level settings. 
This should help creating new task instead of adding everything in one uber-task.

This allows to add separate task UnityRunTests which implements new Unity Test Framework.
Everything should be back compatible now.

PR #1195, opened 2023-06-02
Update Unity BuildTargets enum

kkopanski:
Updated `UnityBuildTarget` enum which has some unknown by Unity values provided.
Fixed values are based on every Unity version starting from 2018.x up to 2023.2 (current).

- [x] Is based on my own work


chrisyarbrough:
Should we order the list the same way Unity does in their [source](https://github.com/Unity-Technologies/UnityCsReference/blob/master/Editor/Mono/BuildTarget.cs)?
That would help comparing that both have the same content.

And I would vote for never removing values from the enum list, even if newer Unity versions remove them, because Nuke can easily support building older legacy Unity versions.

kkopanski:
+1 for not removing.
I will fix it.

I'm not sure about keeping this enum in sync with Unity source.
This will change pretty fast and some values are just deleted sometimes.

As an alternative we could consider adding some extensions methods and helper enums in Unity version namespace.
This could help using only supported values in specified version, for example:

```
namespace Nuke.Common.Tools.Unity2023_1;

public class UnityBuildTarget : Enumeration
{
    public static Unity.UnityBuildTarget Android = Unity.UnityBuildTarget.Android;
    public static Unity.UnityBuildTarget Bratwurst = Unity.UnityBuildTarget.Bratwurst;
    public static Unity.UnityBuildTarget CloudRendering = Unity.UnityBuildTarget.CloudRendering;
    ...
}
```

kkopanski:
Updated BuildTargets with possibilities up to 2023.2.
Old values kept for backward compatibility.

PR #1207, opened 2023-06-26
Add solution source generator path from attribute support and exception reporting

### LINKED ISSUE #1205: Solution source generator can't handle solution path from attribute / does not report exceptions properly
### Usage Information

NUKE 7.0.2 / SDK 7.0.304 / net6.0 / Win10 x64

### Description

There are two closely related issues:

1. The solution source generator does not work when using the `relativePath` constructor parameter of the `SolutionAttribute` instead of defining the solution name in `parameters.json`.
[This](https://github.com/nuke-build/nuke/blob/develop/source/Nuke.SourceGenerators/StronglyTypedSolutionGenerator.cs#L47) part of the code only handles reading the `parameters.json` and subsequently throws when the key is not found.

3. Exceptions in the solution source generator are not caught and reported through the provided source generator logging facilities.
As a result the actual error is swallowed an only an unhelpful compiler error is shown, see "Actual Behavior"

I can provide a pull request that addresses both issues.

### Reproduction Steps

1. Clone NUKE repo
2. Remove `"Solution": "nuke-common.sln",` from `parameters.json`
3. Replace 
`[Solution(GenerateProjects = true)] readonly Solution Solution;`
with
`[Solution("nuke-common.sln", GenerateProjects = true)] readonly Solution Solution;`
in `build\Build.cs`
4. Run `nuke --help` or `dotnet build` in `build` project

### Expected Behavior

Compilation should succeed.

### Actual Behavior

Compilation breaks with opaque error message. 

```
C:\Program Files\dotnet\sdk\7.0.304\Roslyn\Microsoft.CSharp.Core.targets(80,5): error MSB6006: "csc.dll" exited with code 1. [D:\downloads\nuke\build\_build.csproj]
```

### Regression?

Not from what I can tell.

### Known Workarounds

Use solution name from `parameters.json`

Yes

jwfx:
Adds the missing support for supplying the solution name via attribute parameter instead of parameter.json while having source generation enabled, e.g.
`[Solution("my.sln", GenerateProjects = true)] readonly Solution Solution;`

Prevent source generator exceptions from being swallowed and replaced by unhelpful generic compiler errors by using the proper reporting facility `context.ReportDiagnostic()`.

Fixes #1205

- [x] Is based on my own work


matkoch:
Can you please split this in two PRs, or at least two commits? The current one does two things at the same time.

For the exception handling I suggest introducing another `Execute` overload that takes the compilation. This way, the error handling is a single response of the outer method, and there's not a huge indented block as it is now.

jwfx:
> Can you please split this in two PRs, or at least two commits? The current one does two things at the same time.

Done, split it in two commits.

> For the exception handling I suggest introducing another `Execute` overload that takes the compilation. This way, the error handling is a single response of the outer method, and there's not a huge indented block as it is now.

I turned the source generation into a method. It felt a bit awkward to make an overload that would take the context and the compilation from the same context. The context would still have been required because of the [AddSource()](https://github.com/nuke-build/nuke/blob/develop/source/Nuke.SourceGenerators/StronglyTypedSolutionGenerator.cs#L62) call.

matkoch:
This is perfect! Just as I would've done it.

PR #1245, opened 2023-08-30
Fix appveyor configuration generation

olsh:
As per the [documentation](https://www.appveyor.com/docs/packaging-artifacts/) 
> IMPORTANT! If the artifact path starts with *, you need to surround the value with single quotes

- [x] Is based on my own work


matkoch:
Could you refactor to use `SingleQuoteIfNeeded` ?

I don't think we need any comments either.

olsh:
> Could you refactor to use `SingleQuoteIfNeeded` ?
> 
> I don't think we need any comments either.

Done

matkoch:
Not exactly what I meant, but I've fixed it.

PR #1254, opened 2023-09-09
Added Docker CliSettings context and host, along with extensions

fredericvaillancourt:
Changes
- Added Docker host parameter list (CliSettings for all commands)
- Added Docker context parameter (CliSettings for all commands)
- Set CliSettings extensionMethods to true so it can be configured. Otherwise there is no way to set cli settings values ...

Removing the ProcessLogger from CliSettings could have done the trick, to avoid NewInstance throwing while copying the logger. However it seems the exception there was on purpose, to prevent people from using the logger in the common type. So I modified a little bit the tool code generation to support not deriving from ToolSettings, which removes the extra ProcessLogger and makes NewInstance not try to copy anything. I think it makes everything cleaner, not sure what you had in mind at the beginning ...

It was partially tested in our solution, I was able to remove all hacks to set CliSettings properties. I was not able to do a full build though as our CI is currently broken ...

- [X] Is based on my own work

### DIFF
--- a/source/Nuke.Common/Tools/Docker/CliSettings.cs
+++ b/source/Nuke.Common/Tools/Docker/CliSettings.cs
@@ -14,6 +14,4 @@ public partial class CliSettings
     {
         return ConfigureProcessArguments(new Arguments());
     }
-
-    public override Action<OutputType, string> ProcessLogger => base.ProcessLogger ?? throw new NotSupportedException();
 }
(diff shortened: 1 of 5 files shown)


IT-VBFK:
@fredericvaillancourt Can you fix this in Fallout?

fredericvaillancourt:
@IT-VBFK Added a comment on your PR in fallout to figure out if we should commit the generated or not.

Also, the way tools work changed quite a bit. We do not need most of the code I wrote years ago. I added the two missing parameters to the docker json and tried it, looks like they work just fine.

PR #1261, opened 2023-09-15
Fix AbsolutePath attributes matching using GetFiles and GetDirectories

kkopanski:
This will fix usage of `AbsolutePath.GetFiles` and  `AbsolutePath.GetDirectories` with `attributes` argument different than zero.

There was an incorrect check to zero leading to omit files/dirs with specified attributes.
Without fix filter was negative: was excluding entries. 

- [x] Is based on my own work


matkoch:
Thanks! Nice catch.

PR #1298, opened 2023-11-29
Support for actions/checkout@v4 with progress and filter

lahma:
* GH actions/checkout updated to version 4
* Added support for [supplying progress](https://github.com/actions/checkout/pull/1067)
* Added support for [partial checkout filters](https://github.com/actions/checkout/pull/1396)

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common.Tests/CI/ConfigurationGenerationTest.Test_testName=detailed-triggers_attribute=GitHubActionsAttribute.verified.txt
+++ b/source/Nuke.Common.Tests/CI/ConfigurationGenerationTest.Test_testName=detailed-triggers_attribute=GitHubActionsAttribute.verified.txt
@@ -53,11 +53,13 @@ jobs:
       group: ${{ github.workflow }} @ ${{ github.event.pull_request.head.label || github.head_ref || github.run_id }}
       cancel-in-progress: true
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           submodules: recursive
           lfs: true
           fetch-depth: 2
+          progress: false
+          filter: tree:0
       - name: 'Cache: .nuke/temp, ~/.nuget/packages'
         uses: actions/cache@v4
         with:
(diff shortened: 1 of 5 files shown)


Shepard777 (commented):
Hi

Orgl9l:
@lahma - if you're planing on completing this pull request, please can you add in the ability to specificity a `path:` as part of the `with:` to `action@checkout`!

My use case is to check out multiple repos side-by-side per [this](https://github.com/actions/checkout#checkout-multiple-repos-private) However to achieve this I need to check out the base/standard repo to a parallel sub-path.

Such that the block look similar to this:

```yml
      - uses: actions/checkout@v4
        with:
          path: MySpecifiedBaseRepoPath
```

As past of this one also need to fix-up `BuildCmdPath` in  `GitHubActionsAttribute` such that Run step with invoke the launch script `build.cmd` from that sub folder to find the Nuke project!

White I can create an sub class of `GitHubActionsAttribute` and override `protected GitHubActionsJob GetJobs(GitHubActionsImage image, IReadOnlyCollection<ExecutableTarget> relevantTargets)` to add additional steps - in this case to make additional steps to use `actions/checkout@v4`'s for additional repos, it seem that I can't replace the existing `GitHubActionsCheckoutStep` as the underlying attribute fields are `private` and the Property getters are also locked down.

In the ideal case, it would be great to support more of the  `action@checkout` properties, and have built-in support to perform additional checkouts for other repos that might be needed for the build including being able to specific the additional repos SSH secret (deploy key) such that you get blocks similar to this (the assumption is that both repos will have a similar branch):

```yml
      - name: checkout additional repo Org/AbcRepo
        uses: actions/checkout@v4
        with:
          repository: Org/AbcRepo
          ref: ${{ github.ref_name }}
          fetch-depth: 0
          ssh-key: ${{ secrets.REPO_SSH_SECRET }}
          path: MySpecifiedPath
```

I also hope that @matkoch is supportive of this addition.

Thanks!

leorg99:
@lahma @matkoch  What is left to do so we can have this PR approved? I can help work on it. Is it just the one failing check and merge conflict?

PR #1299, opened 2023-11-29
Add separate GH Actions root level concurrency group

lahma:
Different image jobs can be cancelling each others when having same job group definition (it's valid to have it on job level too though). Adding a separate concurrency group to root level will cancel the entire workflow including it's jobs instead of jobs fighting against each other will make the default multi-image setup easier.

- [x] Is based on my own work


Shepard777:
Hey

Shepard777 (commented):
Hey

PR #1337, opened 2024-02-18
Use json file directly in bootstrap

### LINKED ISSUE #1336: Remove `perl` dependency from unix bootstrap and use global.json directly
### Description

The default bootstrap scripts should set the `-JSonFile|--jsonfile` parameter on the `dotnet-install` script if the `global.json` file exists. The main motivation for this is to remove `perl` as a dependency on the linux side.

### Usage Example

_No response_

### Alternative

Everyone who is interested in this can manually edit the bootstrap scripts and make it work differently

Yes

KuraiAndras:
This PR fixes #1336.

This PR also introduces a breaking change. When the consumer has a `global.json` file, but it does not have an `sdk` property with `version` set then the current behavior is to fall back to using the `STS` channel. In that case, this will cause the `dotnet-install` script to fail. Is this okay for a breaking change? Is this something that was considered when designing the current dotnet version resolution?

- [X] Is based on my own work

### DIFF
--- a/build.ps1
+++ b/build.ps1
@@ -22,6 +22,8 @@ $DotNetChannel = "STS"
 
 $env:DOTNET_CLI_TELEMETRY_OPTOUT = 1
 $env:DOTNET_NOLOGO = 1
+$env:DOTNET_ROLL_FORWARD = "Major"
+$env:NUKE_TELEMETRY_OPTOUT = 1
 
 ###########################################################################
 # EXECUTION
--- a/build.ps1
+++ b/build.ps1
@@ -1,6 +1,6 @@
 [CmdletBinding()]
 Param(
-    [Parameter(Position=0,Mandatory=$false,ValueFromRemainingArguments=$true)]
+    [Parameter(Position = 0, Mandatory = $false, ValueFromRemainingArguments = $true)]
     [string[]]$BuildArguments
 )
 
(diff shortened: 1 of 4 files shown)


matkoch:
Where does `DotNetVersion` come from now?

And I didn't quite get what you mean about breaking change. Why does `-Channel STS` fail?

build.ps1
```diff
@@ -22,8 +22,6 @@ $DotNetChannel = "STS"
 
 $env:DOTNET_CLI_TELEMETRY_OPTOUT = 1
 $env:DOTNET_NOLOGO = 1
-$env:DOTNET_ROLL_FORWARD = "Major"
```
matkoch: revert please

KuraiAndras:
Well, `DotNetVersion` is never declared now as a variable inside the build scripts. I thought that if you want to use that you either declare it in your own build script or provide it through an environment variable?

On the breaking change:
In the current version if your global.json looks like this for example:
```json
{
  "sdk": {
    "allowPrerelease": false
  }
}
```
Meaning that it does not contain the `version` property, the current behavior is to use the value of `DotNetChannel`.

With this change if you have a `global.json` file **and** you don't provide the `version` property the file will be passed to the dotnet install script and it will fail. https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script#options

matkoch:
Gotcha. I think in that case it doesn't make much sense to chance the implementation where it potentially breaks. In contrast, I've yet to see a bug report about the Perl dependency. Back then, I specifically researched how well it is supported among Linux distros.

I actually discussed the `--jsonfile` approach with someone else recently. But now it seems that it's not a good idea. Once the .NET team supports a missing `version` property, we can reevaluate this.

matkoch:
Thanks anyway for the effort!

PR #1349, opened 2024-03-14
Force EnableUnsafeBinaryFormatterSerialization using SetSwitch - BinaryFormatter Issue

ricaun:
Hello everyone,

This PR forces to `EnableUnsafeBinaryFormatterSerialization` in runtime instead of adding the configuration in the csproj. 

This is done by using the code below before using the `BinaryFormatter`. (Reference: https://github.com/pythonnet/pythonnet/issues/2282)

```c#
AppContext.SetSwitch("System.Runtime.Serialization.EnableUnsafeBinaryFormatterSerialization", true);
```

I add a simple test in the Nuke.Tooling.Tests and changes to NET 8.0 to make sure the `BinaryFormatter` gonna work without any `EnableUnsafeBinaryFormatterSerialization` in the csproj.

This gonna make sure the `BinaryFormatter` gonna works regardless if has or hasn't the configuration in the csproj.

And just to remind that the`BinaryFormatter` gonna be removed in the next .NET Core (.NET 9.0 - https://github.com/dotnet/announcements/issues/293)

---

- [x] Is based on my own work

### DIFF
--- a/source/Directory.Build.props
+++ b/source/Directory.Build.props
@@ -8,7 +8,7 @@
     <DebugType>embedded</DebugType>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
-    <EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>
+    <!--<EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>-->
   </PropertyGroup>
 
   <PropertyGroup>
--- a/source/Directory.Build.props
+++ b/source/Directory.Build.props
@@ -8,7 +8,6 @@
     <DebugType>embedded</DebugType>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
     <EmbedUntrackedSources>true</EmbedUntrackedSources>
-    <!--<EnableUnsafeBinaryFormatterSerialization>true</EnableUnsafeBinaryFormatterSerialization>-->
   </PropertyGroup>
 
   <PropertyGroup>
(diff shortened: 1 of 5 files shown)


matkoch:
What exactly is the benefit compared to the auto-loaded `.props` file?

ricaun:
The `.props` is goes only to the main `csproj` that referencing the `Nuke.Commum`.

This mean if I create a package with the dependencies `Nuke.Commum` or `Nuke.Tooling`, the `.props` does not progate to the main `csproj` that use the package.

Here is my package the uses that. https://www.nuget.org/packages/ricaun.Nuke#dependencies-body-tab

Now I could add the `.props` in my `ricaun.Nuke` package, or try to fix in the `Nuke.Tooling` to force enable the switch on runtime fixing the issue with net8.0.

The benefits would be using the `Nuke.Tooling` without any extra confuguratuon in the `csproj` or `.props`.

ITaluone:
This looks like the proper solution (especially for projects which using/extending Nuke.*)

ricaun:
At the moment I gonna use this in my package.

* https://github.com/ricaun-io/ricaun.Nuke/blob/develop/ricaun.Nuke/Components/Module.cs

By default when the DLL is load the module switch the `EnableUnsafeBinaryFormatterSerialization` to true.

ScarletKuro:
Sorry for necro posting, but setting
https://github.com/nuke-build/nuke/blob/abeb8afc9ec80abf1c92e14549b3b8b2d00bae8f/source/Nuke.Tooling/SettingsEntity.NewInstance.cs#L21
Is not enough, at least for the net9 SDK as the code for binary formatter is missing there.
You also need to include the package
```xml
<ItemGroup>
    <PackageReference Include="System.Runtime.Serialization.Formatters" Version="9.0.0" />
</ItemGroup>
```

ricaun:
@ScarletKuro adding `System.Runtime.Serialization.Formatters` in .NET 9 make binary formatter to work?

ScarletKuro:
> @ScarletKuro adding `System.Runtime.Serialization.Formatters` in .NET 9 make binary formatter to work?

Yes.
- https://learn.microsoft.com/en-us/dotnet/standard/serialization/binaryformatter-migration-guide/compatibility-package
- https://youtu.be/4iEqqPZKDC0?si=QoRJBxFLAdAx4nhU&t=1720

MSFT talked about the removal of BinaryFormatter from .NET 9 a long time ago and providing compatibility package.

PR #1356, opened 2024-03-25
Fix package source name

werwolfby:
Default global `nuget.config` uses `nuget.org` as key.

In my case this global `nuget.config` has private source, and special `packageSourceMapping` to use my private source for some packages. And use `nuget.org` as default one for everything else:

```xml
  <packageSourceMapping>
    <packageSource key="nuget.org">
      <package pattern="*" />
    </packageSource>
    <packageSource key="werwolf.by">
      <package pattern="werwolfby.*" />
    </packageSource>
  </packageSourceMapping>
```

Because `nuget.config` specified `api.nuget.org` as default, my global one doesn't allow me to restore packages in `nuke` project. This fix returns everything how it should be.

There is the question if `nuke` requires `nuget.config` at all?

- [x] Is based on my own work

### DIFF
--- a/NuGet.config
+++ b/NuGet.config
@@ -4,6 +4,6 @@
     <add key="disableSourceControlIntegration" value="true" />
   </solution>
   <packageSources>
-    <add key="api.nuget.org" value="https://api.nuget.org/v3/index.json" />
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" />
   </packageSources>
 </configuration>


gtbuchanan:
> There is the question if nuke requires nuget.config at all?

I'll second this question. I also ran into this issue since I have package source mapping configured in my user-level `NuGet.Config`. I deleted `NuGet.Config` locally from the Nuke repository to move forward with my PR, but I suppose this rename would have worked as well.

matkoch:
simply renaming doesn't truly fix the issue. the next person could be using yet another name. it requires a `<clear />` tag.

PR #1363, opened 2024-04-11
Add ability to specify labels and group for self-hosted runners

leorg99:
Added ability in Github Actions to specify [groups and labels to route jobs](https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/using-self-hosted-runners-in-a-workflow#using-labels-and-groups-to-route-jobs) with a self-hosted runner

- [x] Is based on my own work


Simonl9l:
@matkoch seems @leorg99 is onto something...given where you are here with merging this PR in, are there any consideration to supporting custom runners images - in our case were trying to build AoT binaries and seem easiest to define an ARM runner for each platform. It seem like the `runs-on:`, `labels:` seem to cover it?

If there is anything we can do to help in this regard please advise.

Issue #1173, opened 2023-05-15
Support Azure Devops self-hosted agents


onybo:
### Description

I would really like to see support for [Self-hosted agents](https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/agents?view=azure-devops&tabs=browser#install) in addition to the current  [Microsoft hosted agents](https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/agents?view=azure-devops&tabs=browser#microsoft-hosted-agents).

I am aware that a [similar issue](https://github.com/nuke-build/nuke/issues/1019) was raised some time ago, but in the previous issue it was more about on-prem agents while I think it is really about supporting the more general concept of self-hosted agents, which may be on-prem or anywhere else.

It seems that the current implementation of AzurePipelines in nuke is only for [Microsoft hosted agents](https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/agents?view=azure-devops&tabs=browser#microsoft-hosted-agents).
As far as I know, the generated yaml only works when using Microsoft hosted agents and the AzurePipelines attribute can't be configured so that the yaml file required for a Self-hosted agent is generated.

Furthermore the generated yaml is assuming that all jobs/targets should go into all stages, and each stage is linked to a specific vmImage.
E.g. adding an AzurePipelines attribute like this will produce two almost identical stages including all targes as jobs (only difference would be the vm image):

```
[AzurePipelines( 
    AzurePipelinesImage.UbuntuLatest,
    AzurePipelinesImage.Ubuntu1804,
    InvokedTargets = new[] { nameof(Compile) })]
```
Producing a yaml wich would look like this:
![image](https://github.com/nuke-build/nuke/assets/1699082/3d93997c-4d7c-4823-93f5-81d846f4320a)

Running all targets (jobs) on all images may make sense when all agents have the same set of installed tools like the Microsoft hosted agents, but when using self hosted agents this may not make as much sense.

When using self hosted agents we should not be specifying the `vmImage` property of the `pool`, but rather the name of the `pool` that we want to use.
When self hosted agents they may also have different capabilities and we need to control what `pool` is used for a specific target/job.
E.g. the test targets/jobs may require different tools than what is needed for the build job.

I suggest to introduce a new `AzureStageAttribute` which could accept either a stage name or the image. 
As we can have any combination of Stages running on either Microsoft hosted and some on self-hosted I think we should allow for multiple `AzureStageAttributes` on the Build object. 
[...]

Issue #749, opened 2021-04-27
AzurePipelinesAttribute - allow to set `pool name` from attribute


rafek1241:
### Description

```
pool:
  name: 'XXX' <- this one cannot be set at all
  vmImage: 'ubuntu-latest' <-- this one can be set currently via AzurePipelinesImage enum
```

### Usage Example

```code
[AzurePipelines(
    AzurePipelinesImage.UbuntuLatest,
    PoolName = "elo" //<- new
    AutoGenerate = true,
    InvokedTargets = new[] { nameof(ContinousIntegration) },
    TriggerBranchesExclude = new[] { "master" }
)]
```

### Alternative

None

matkoch:
@rafek1241 have you read the issue template?

rafek1241:
@matkoch 
Yes, I read about sponsorship requirement (are you talking about that one?). I will do this in spare time.

matkoch:
what do you mean spare time?

rafek1241:
Oh, my bad. I mean I will do this change after working hours.

matkoch:
Unfortunately that doesn't change the fact :) It's about my time.

guillaume86:
@matkoch forgive me if I'm missing something obvious, but I don't really get what happened here, can you explain?
I looked at the issue template for feature suggestions and did not see anything about sponsorships.
I'm evaluating nuke for my project but I would need this feature if I decide to go ahead, I could probably submit a PR if it's OK'd.

matkoch:
I've relaxed the rules a bit, but I'm still very occupied with a lot of other things currently. So definitely take this into account.

guillaume86:
Ha ok, thanks for your reply I'll keep this in mind.

PR #1398, opened 2024-07-12
- added AzurePipelinesPool as a abstraction for vmImage and on prem p…

mateuszradny:
I want to use nuke.build with our on-prem agents. Unfortunately, I didn't find a solution to pass a pool name to be considered in the final yml file.

I saw that other people also had a similar problem: #1173, #749.

My proposition is to create an additional abstraction, AzurePipelinesPool, which can be either a Microsoft-hosted agent (vmImage) or a self-hosted agent, as that is essentially what a pool is in Azure Pipelines.

Here is a proposal for the changes, and I am open to suggestions. 
I also removed `images` parameters from `AzurePipelinesAttribute` because if you really need to run your targets on multiple vmImages, it's probably better to configure multiple `AzurePipelinesAttribute`, am I right?

- [x] Is based on my own work

### DIFF
--- a/.nuke/build.schema.json
+++ b/.nuke/build.schema.json
@@ -46,14 +46,8 @@
       "type": "string",
       "default": "Secrets must be entered via 'nuke :secrets [profile]'"
     },
-    "SignPathOrganizationId": {
-      "type": "string"
-    },
-    "SignPathPolicySlug": {
-      "type": "string"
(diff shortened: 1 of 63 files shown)


MikeFarrington:
How is this looking?  Any chance we can see it in 9.0.5?

PR #1405, opened 2024-08-03
Update dotnet run command

### LINKED ISSUE #1403: dotnet run does not support /property MSBuild parameters
### Usage Information

NUKE 8.0.0 / SDK 8.0.303 / net8.0 / win-x64

### Description

Adding MSBuild properties to DotNetTasks.DotNetRun() calls currently produces `/property` parameters, which is not supported, it should be `--property`.

`dotnet run doesn't respect arguments like /property:property=value, which are respected by dotnet build.`
Source: https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-run#description

### Reproduction Steps

```
DotNetTasks.DotNetRun(y => y
    .SetApplicationArguments("--someparameter")
    .SetProjectFile(RootDirectory / "test" / "test.csproj" )
    .AddProperty("VersionPrefix", "9.0.0"));
```

### Expected Behavior

`14:22:44 [INF] > "C:\Program Files\dotnet\dotnet.exe" run --project D:\test\test\test.csproj --property:VersionPrefix=9.0.0 -- --someparameter`

### Actual Behavior

`14:22:44 [INF] > "C:\Program Files\dotnet\dotnet.exe" run --project D:\test\test\test.csproj /property:VersionPrefix=9.0.0 -- --someparameter`

### Regression?

_No response_

### Known Workarounds

_No response_

No

jwfx:
Fixes #1403

Main goal was to get `AddProperty()` working, as `dotnet run` requires `--property` parameters instead of `/property`.

Added all missing parameters to `dotnet run`, while taking parameters from `dotnet run --help` of .NET SDK 8.0.303 as source of truth.

There is some inconsistency between [dotnet run docs](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-run) and `dotnet run --help`. The docs have parameters (e.g. `--force`) that `--help` is not showing and it in turn shows parameters like `--self-contained` which are not part of the docs.

Let me know if I made the wrong call here.

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common/Tools/DotNet/DotNet.json
+++ b/source/Nuke.Common/Tools/DotNet/DotNet.json
@@ -203,10 +203,22 @@
       "settingsClass": {
         "properties": [
           {
+            "name": "Architecture",
+            "type": "string",
+            "format": "--arch {value}",
+            "help": "Specifies the target architecture. This is a shorthand syntax for setting the <a href=\"https://learn.microsoft.com/en-us/dotnet/core/rid-catalog\">Runtime Identifier (RID)</a>, where the provided value is combined with the default RID. For example, on a win-x64 machine, specifying --arch x86 sets the RID to win-x86. If you use this option, don't use the -r|--runtime option. Available since .NET 6 Preview 7."
+          },
+          {
             "name": "Configuration",
             "type": "string",
             "format": "--configuration {value}",
             "help": "Configuration to use for building the project. The default value is Debug."
+          },
+          {
+            "name": "DisableBuildServers",
+            "type": "bool",
+            "format": "--disable-build-servers",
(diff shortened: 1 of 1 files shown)


matkoch:
That's a whole bunch of duplication. Why not change `/property` here?

https://github.com/nuke-build/nuke/blob/06e8e17f17b70fc6c81eecdcf7cfb03c08a1fc0b/source/Nuke.Common/Tools/DotNet/DotNet.json#L1701

jwfx:
Initially I wasn't sure all relevant dotnet commands would actually support the `--property` syntax, because of the spotty documentation on MS side.

I took the time to verify it for
build
clean
pack
publish
restore
run
test

So I made the change for commonTaskProperties as per your suggestion.

It looks like we still need the duplication though, because  `dotnet msbuild` does not support this syntax.

PR #1421, opened 2024-09-13
Update 07-solution-project-model.md

PieEatingNinjas:
I could not find the 'GetProjects' method. I think it should be 'GetAllProjects'

- [x] Is based on my own work


BrunoVT1992:
I also saw this as an issue and solution

Issue #1436, opened 2024-10-23
Full interactive mode (ask for target)


Danielku15:
### Description

Nuke is great and very flexible when it comes to CLI usage and is very flexible in the usage from CI/CD systems. What I am missing a bit, is a modern user experience when developers use the "_build" project locally. 

If the project/build is started without any input, and we detect a "Terminal" host, we enter an interactive mode to ask the user what to do. e.g. we have 5 different entry points to our build depending on what you want to do. 

### Usage Example

When I launch the "_build.exe" (or dotnet run) Nuke should prompt the user:

```
​
███╗   ██╗██╗   ██╗██╗  ██╗███████╗
████╗  ██║██║   ██║██║ ██╔╝██╔════╝
██╔██╗ ██║██║   ██║█████╔╝ █████╗  
██║╚██╗██║██║   ██║██╔═██╗ ██╔══╝  
██║ ╚████║╚██████╔╝██║  ██╗███████╗
╚═╝  ╚═══╝ ╚═════╝ ╚═╝  ╚═╝╚══════╝
​
NUKE Execution Engine version 8.1.2 (Windows,.NETCoreApp,Version=v8.0)
​
Targets (with their direct dependencies):

  > PrepareEnviroment
      Checks for any missing environment configurations on your machine and sets up whatever is needed.
    Build
      Compile the whole project.
    Test
      Run the tests of this project 

Select which target you want to run with the arrow keys and run it with [ENTER].
```

Then it starts the respective target just as if `nuke --target PrepareEnvironment` is called. The way of asking could be either fully interactive like very modern console apps, or simply asking for the user to type it. 

```
​
███╗   ██╗██╗   ██╗██╗  ██╗███████╗
████╗  ██║██║   ██║██║ ██╔╝██╔════╝
██╔██╗ ██║██║   ██║█████╔╝ █████╗  
██║╚██╗██║██║   ██║██╔═██╗ ██╔══╝  
██║ ╚████║╚██████╔╝██║  ██╗███████╗
╚═╝  ╚═══╝ ╚═════╝ ╚═╝  ╚═╝╚══════╝
​
NUKE Execution Engine version 8.1.2 (Windows,.NETCoreApp,Version=v8.0)
​
Targets (with their direct dependencies):

  PrepareEnviroment
    Checks for any missing environment configurations on your machine and sets up whatever is needed.
  Build
    Compile the whole project.
  Test
    Run the tests of this project 

[...]

PR #1437, opened 2024-10-24
feat: Add Interactive mode with target selection

Danielku15:

- [x] Is based on my own work

---

This PR contains a proposal how https://github.com/nuke-build/nuke/issues/1436 could be implemented. Structurally we should still discuss where what should go but it already gives a feeling on what a user gets with this change. 

### Current design decisions

* I opted to the SpecterConsole as proposed in the issue. 
* I moved the `InvokedTarget` loading logic into the `ExecutableTargetFactory`
* I put the implementation of selecting targets into `NukeBuild` to allow people overriding and customizing it. 
* I made the `ExecutableTargets` public so custom implementations can use it. 
* For the sake of simplicity i did not move yet the console logic into the `Terminal` class. 
* I tried to keep the target selection output close to what a user sees when executing `--help`

Any feedback is welcome to complete this change.

### Demo

https://github.com/user-attachments/assets/1a59d2be-1444-4b9e-9361-7d0a788f0d61

```cs
class Build : NukeBuild
{
    public static int Main() => Execute<Build>();

    public Build()
    {
        Interactive = true;
    }

    public Target Compile => t => t
        .Description("Compiles the project")
        .Executes(() => Log.Information("Compile target execute"));

    public Target Test => t => t
        .DependsOn(Compile)
        .Description("Executes all tests")
        .Executes(() => Log.Information("Test target execute"));

    public Target PrepareEnvironment => t => t
        .Before(Compile)
        .Description("Prepares your development environment with all dependencies.")
        .Executes(() => Log.Information("PrepareEnvironment target execute"));
}
```

### DIFF
--- a/source/Nuke.Build/INukeBuild.cs
+++ b/source/Nuke.Build/INukeBuild.cs
@@ -35,7 +35,7 @@ public interface INukeBuild
     IReadOnlyCollection<ExecutableTarget> SucceededTargets { get; }
     IReadOnlyCollection<ExecutableTarget> FinishedTargets { get; }
 
-    bool IsSuccessful { get; }
+    bool IsSucceeding { get; }
     bool IsFailing { get; }
     bool IsFinished { get; }
     int? ExitCode { get; set; }
--- a/source/Nuke.Build/INukeBuild.cs
+++ b/source/Nuke.Build/INukeBuild.cs
@@ -19,7 +19,7 @@ public interface INukeBuild
 {
     void ReportSummary(Configure<Dictionary<string, string>> configurator = null);
 
-    internal IReadOnlyCollection<ExecutableTarget> ExecutableTargets { get; }
+    IReadOnlyCollection<ExecutableTarget> ExecutableTargets { get; }
     internal IReadOnlyCollection<IBuildExtension> BuildExtensions { get; }
     internal bool IsInterceptorExecution { get; }
     internal string[] LoadedLocalProfiles { get; }
@@ -52,6 +52,7 @@ public interface INukeBuild
     bool Plan { get; }
     bool Help { get; }
     bool NoLogo { get; }
+    bool Interactive { get; }
     bool IsLocalBuild { get; }
     bool IsServerBuild { get; }
     bool Continue { get; }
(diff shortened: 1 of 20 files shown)


source/Nuke.Build/INukeBuild.cs
```diff
@@ -19,7 +19,7 @@ public interface INukeBuild
 {
     void ReportSummary(Configure<Dictionary<string, string>> configurator = null);
 
-    internal IReadOnlyCollection<ExecutableTarget> ExecutableTargets { get; }
+    IReadOnlyCollection<ExecutableTarget> ExecutableTargets { get; }
```
matkoch, line 22: Should be reverted

kusc-leica: From my side this change was intentional. To properly be able to override and customize `OnNoTargetsSpecified()` developers need access to the general list of targets which are there as option. 

e.g. in some projects we'd like to only be able to have a "single target" selection as this fits the project setup better and would like to change the default "multi selection" implementation shipped in nuke. 

As alternative I could pass the executableTargets a parameter to `OnNoTargetsSpecified`.

Danielku15: From my side this change was intentional. To properly be able to override and customize `OnNoTargetsSpecified()` developers need access to the general list of targets which are there as option. 

e.g. in some projects we'd like to only be able to have a "single target" selection as this fits the project setup better and would like to change the default "multi selection" implementation shipped in nuke. 

As alternative I could pass the executableTargets a parameter to `OnNoTargetsSpecified`.

ITaluone:
I wouldn't go for "automatically detect local terminal and execute the interactive mode when no parameters given".

I would rather add an option on `nuke :setup` as opt-in. Maybe not everyone wants to be forced into interactive mode, especially, when running `.\build.cmd` before runs the pipeline and suddenly it stays in interactive mode, waiting for user input.
(Can also be considered as a behavioral breaking change IMHO)

Danielku15:
> I wouldn't go for "automatically detect local terminal and execute the interactive mode when no parameters given".

@ITaluone I agree and there are some considerations around that in the proposal. One important detail in this implementation is: there is an additional `--interactive` parameter (or `bool Interactive {get;set;}`) which is default unset. With the current implementation you have following options as a developer/user:

* You set `--interactive` on the commandline to activate the mode. 
* You can set `this.Interactive = <your own check>` in the constructor of your Build class to control things. 
* You can override the `OnNoTargetsSpecified` method and trigger any custom logic.

PR #1486, opened 2024-12-04
Fix `TaskGenerator` tool path with environment variable

ricaun:
Fix `TaskGenerator` to use `ToolTasks.GetToolPath<T>()` to enable use environment variable like nuke `8.*`.

Related to this PR: #1485

Here is a before and after what change in the `.Generated.cs` files, I add a summery to show that the environment variable is used.

### Before
```C#
public static string AzureSignToolPath => new AzureSignToolTasks().GetToolPath();
```

### After
```C#
/// <summary>Get the AzureSignTool executable path or 'AZURESIGNTOOL_EXE' environment variable.</summary>
public static string AzureSignToolPath => ToolTasks.GetToolPath<AzureSignToolTasks>();
```

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Tooling.Generator/Generators/TaskGenerator.cs
+++ b/source/Nuke.Tooling.Generator/Generators/TaskGenerator.cs
@@ -1,4 +1,4 @@
-// Copyright 2023 Maintainers of NUKE.
+﻿// Copyright 2023 Maintainers of NUKE.
 // Distributed under the MIT License.
 // https://github.com/nuke-build/nuke/blob/master/LICENSE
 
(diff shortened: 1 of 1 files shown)


matkoch:
appreciate the PR in this dark territory, but i'd prefer to fix this in a slightly different way... will be released soon

ricaun:
> appreciate the PR in this dark territory, but i'd prefer to fix this in a slightly different way... will be released soon

Great! 
Was fun to figure out how the tool generator works, I was using version `8.*`  so was easier to understand and unit test using the environment variable.

ricaun:
I like your solution, now is possible to set the `ToolPath` without messing with environment variable! 

Nice! 👏

matkoch:
Thanks! And instance methods so that code completion doesn't show non-user methods.

PR #1505, opened 2025-01-16
Add GitHub actions/setup-dotnet support

lahma:
More often than I want, I seem to hit the oddities of different GH runner images. sometimes they have both NET 8 and NET 9, sometimes only either one. NUKE's setup seems to force the one in global JSON and then when some component explicitly would want to use the other version it's missing (because installation replaces the older one - if installing 9, 8 will be removed).

This PR adds support for specifying versions of .NET that should all be present.

- [x] Is based on my own work


matkoch:
DM :)

PR #1506, opened 2025-01-18
Update Build.cs

coendewit:
Nuke.Common.IO.FileSystemTasks does not exist and broke my first time use.

When using the latest version 9.0.4 on macOS, I could not follow the first baby steps: install, setup and run. The setup template referenced a non existent FileSystemTasks. I don't know where it was moved to, but since it is not used in code anyway, I propose to at least remove it.


clicky:
I can confirm I had this issue as well and was about to create a PR for the same issue with the same solution, thanks @coendewit

PR #1518, opened 2025-02-13
Fix: Add missing ArtifactsPath property to dotnet restore

SigtryggurO:
Add missing ArtifactsPath property to dotnet restore

- [x] Is based on my own work

### DIFF
--- a/source/Nuke.Common/Tools/DotNet/DotNet.json
+++ b/source/Nuke.Common/Tools/DotNet/DotNet.json
@@ -347,6 +347,12 @@
             "type": "DotNetVerbosity",
             "format": "--verbosity {value}",
             "help": "Sets the verbosity level of the command. Allowed values are <c>q[uiet]</c>, <c>m[inimal]</c>, <c>n[ormal]</c>, <c>d[etailed]</c>, and <c>diag[nostic]</c>."
+          },
+          {
+            "name": "ArtifactsPath",
+            "type": "string",
+            "format": "--artifacts-path {value}",
+            "help": "All build output files from the executed command will go in subfolders under the specified path, separated by project. For more information see <a href=\"https://learn.microsoft.com/en-us/dotnet/core/sdk/artifacts-output\">Artifacts Output Layout</a>. Available since .NET 8 SDK."
           }
         ]
       }


ITaluone:
Is this an undocumented argument? I cannot see this on the _restore_ command...

image

MikeFarrington:
> Is this an undocumented argument? I cannot see this on the _restore_ command...

It's not undocumented, but it looks like that page by Microsoft is not version-specific and caters to a least-common denominator of .NET Core 3.

image

This shell command 
```PS
dotnet restore /?
```

Yields a more up to date result (I have the latest SDKs installed):
```
Description:
  .NET dependency restorer

Usage:
  dotnet restore [<PROJECT | SOLUTION>...] [options]

Arguments:
  <PROJECT | SOLUTION>  The project or solution file to operate on. If a file is not specified, the command will search the current directory for one.

Options:
  --disable-build-servers             Force the command to ignore any persistent build servers.
  -s, --source <SOURCE>               The NuGet package source to use for the restore.
  --packages <PACKAGES_DIR>           The directory to restore packages to.
  --ucr, --use-current-runtime        Use current runtime as the target runtime.
  --disable-parallel                  Prevent restoring multiple projects in parallel.
  --configfile <FILE>                 The NuGet configuration file to use.
  --no-http-cache                     Disable Http Caching for packages.
  --ignore-failed-sources             Treat package source failures as warnings.
  -f, --force                         Force all dependencies to be resolved even if the last restore was successful.
                                      This is equivalent to deleting project.assets.json.
  -r, --runtime <RUNTIME_IDENTIFIER>  The target runtime to restore packages for.
  --no-dependencies                   Do not restore project-to-project references and only restore the specified project.
  -v, --verbosity <LEVEL>             Set the MSBuild verbosity level. Allowed values are q[uiet], m[inimal], n[ormal], d[etailed], and diag[nostic].
  --interactive                       Allows the command to stop and wait for user input or action (for example to complete authentication).
  --artifacts-path <ARTIFACTS_DIR>    The artifacts path. All output from the project, including build, publish, and pack output, will go in subfolders under the specified path.
  --use-lock-file                     Enables project lock file to be generated and used with restore.
  --locked-mode                       Don't allow updating project lock file.
  --lock-file-path <LOCK_FILE_PATH>   Output location where project lock file is written. By default, this is 'PROJECT_ROOT\packages.lock.json'.
  --force-evaluate                    Forces restore to reevaluate all dependencies even if a lock file already exists.
  -a, --arch <ARCH>                   The target architecture.
  -?, -h, --help                      Show command line help.
```

It's not just 'restore' that accepts --artifacts-path, other commands like 'build' accept it as well.  DotNetBuildSettings already got the SetArtifactsPath treatment, but DotNetRestoreSettings hadn't.

ITaluone:
Ahh.. I see.. But for `restore` it means: restore to this specific folder instead of `MyProject / obj`.

ITaluone:
@SigtryggurO Do you mind opening this against https://github.com/Fallout-build/Fallout ?

PR #1545, opened 2025-07-22
Update MSBuild.Generated.cs to include properties used in ClickOnce Publication

smsaleem:

- [x] Is based on my own work


VolkmarR:
It looks like you modified a generated file... If you look at the top of the file, it should show the source json file, that contains the definition used to generate it.

smsaleem:
Means MSBuild.json needs an update manually ?

VolkmarR:
I'm non an expert in this, but i know, that there is a process to generate the tools definitions. First you need to update the msbuild.json and then run a nuke target to re-generate the files.

valadas:
In think the target is GenerateTools, so you would run
`nuke --target GenerateTools`

But don't quote me on it, it has been a long time I had to do that...

smsaleem:
Thanks.  It worked.

smsaleem:
A separate [PR](https://github.com/nuke-build/nuke/pull/1546) for MSBuild.json has been created

PR #1555, opened 2025-09-16
Add BuildCompletion to AzurePipelinesBuildReason enum

vmladenov-at-tricentis:
The AzurePipelinesBuildReason has one extra value BuildCompletion

According to the documentation: https://learn.microsoft.com/en-us/azure/devops/pipelines/build/variables?view=azure-devops&tabs=yaml#build-variables-devops-services  the AzurePipelinesBuildReason has one extra property: BuildCompletion

- [x] Is based on my own work


danielmeza:
Shall we able to specify which build will trigger it?

